PR #11, opened 2015-04-18
Updated the README.rst to show the actual output

jieter:
Noticed this while debugging the use of the lib. Double checked it by running the tests: they are all OK.


k-bx:
Nice catch, thanks!

jieter:
;)

PR #16, opened 2015-06-20
Adds max_ver and min_ver APIs. Fixes #15

### LINKED ISSUE #15: Add a max(ver1, ver2) function
Add a max function that compares two semver numbers and returns the larger of the two. Similar to the standard library max when passing in ints. Could also use a min version as well.

zlalanne:
Added max_ver and min_ver APIs that work like the max() and min() of the
standard library when comparing ints. Includes tests and example in
README

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ False
 '3.5.0'
 >>> semver.bump_patch("3.4.5")
 '3.4.6'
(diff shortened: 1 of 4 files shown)


k-bx:
Great, thanks!

PR #18, opened 2015-07-31
Fix a bug with comparing rc0 vs rc1

TalAntR:
Fixed an issue when I'm trying to compare('0.10.1-rc1', '0.10.1-rc0')


k-bx:
Thanks!

PR #20, opened 2016-01-28
Add bumping prerelease and build versions

ppkt:
I added bumping prerelease and build version. It works quite well, I just wasn't sure how semver should behave when we are going to bump prerelease or build on version 1.2.3, so I assume to append 'rc.1' / 'build.1' respectively instead of raising exception.


k-bx:
Looks good, thank you!

k-bx:
Released as 2.3.0

PR #22, opened 2016-02-11
Fix prerelease ordering

### LINKED ISSUE #21: Prerelease comparisons not according to semver.org
It seems that comparisons of pre-release versions do not work according semver.org specification:

Example:
```
$ python -c "import semver; print(semver.min_ver('1.2.3-rc2', '1.2.3-rc10'))"
1.2.3-rc2
```

According to semver.org, the parts after dash should be split at '.', and alphanumeric parts compared lexicographically and numeric parts numerically. It seems that this is not the case in semver-the-module.

tazle:
Fixes #21

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -34,9 +34,9 @@ def parse(version):
 def compare(ver1, ver2):
     def nat_cmp(a, b):
         a, b = a or '', b or ''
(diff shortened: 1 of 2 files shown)


k-bx:
Thanks a lot!

PR #23, opened 2016-03-03
Fix bug in comparision between versions with build and prerelease

ppkt:
I found a bug when versions with RC and build numbers were compared, for example:
```
>>> compare('1.1.9-rc.1', '1.1.9+build.1')
-1
>>> compare('1.1.9-rc.1', '1.1.9-rc.1+build.1')
0
```
It seems like comparision for build values was not implemented, so I added this in my PR :)


k-bx:
Thank you! :+1:  Released under 2.4.1.

PR #25, opened 2016-03-18
(not semver) add two match model for '~' and '^'

chinakids:
add two match model for '~' and '^'

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -13,7 +13,6 @@ _LAST_NUMBER = re.compile(r'(?:[^\d]*(\d+)[^\d]*)+')
 if not hasattr(__builtins__, 'cmp'):
     cmp = lambda a, b: (a > b) - (a < b)
 
-
 def parse(version):
     """
     Parse version to major, minor, patch, pre-release, build parts.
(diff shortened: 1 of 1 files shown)


k-bx:
@chinakids thanks for the contribution. However, I can't accept it as is. First of all, `~` and `^` don't seem like part of a "pure" semantic versioning. Secondly, tests are lacking.

I don't mind adding this npm's advanced ranges syntax, but as a separate API. It's clearly stated [in the docs](https://docs.npmjs.com/misc/semver#advanced-range-syntax) that "Advanced range syntax desugars to primitive comparators in deterministic ways.", which means you can add functions like `match_advanced`, which would reuse current `match` function underneath after desugaring advanced operators.

I'd be happy to accept PR in this case. Thanks.

chinakids:
Thanks

iamale:
I believe that supporting all of the [advanced range syntax](https://docs.npmjs.com/misc/semver#advanced-range-syntax) would be nice. If this needs to be implemented in a separate function, I think `satisfies` is a great choice as this is what it is called in the node semver module.

tomschr:
As Konstantine already said, it should be done in a separate API (class, ...). I've linked the respective issue, so closing this.

PR #26, opened 2016-05-15
Convert README to reStructuredText

bittner:
Currently, the README [does not display correctly](https://pypi.python.org/pypi/semver) on PyPI. 

* Convert README to reStructuredText (as required by PyPI)
* Add more badges


bittner:
Note that the [reference in setup.py](https://github.com/k-bx/python-semver/blob/master/setup.py#L5) to `README.md` also needs to be changed to `README.rst`. This will create a merge conflict with PR #27 though. I suggest we leave it as it is in this PR, and fix it in #27 after this PR is merged.

k-bx:
Thank you!

PR #29, opened 2016-05-23
Support matching 'not equal' with '!='

robi-wan:
Match versions with a match expression meaning 'not equal', for example to explicitly exclude a specific version.

Remove check for unused/unsupported match expression '='.


k-bx:
@robi-wan  Thank you! Released under 2.5.0

PR #30, opened 2016-06-06
Remove comparison of build metadata

jeloagnasin:
According to [Semantic Versioning 2.0.0](http://semver.org/):

> 10. ... Build metadata SHOULD be ignored when determining version precedence. ...
>
> 11. ... Precedence MUST be calculated by separating the version into major, minor, patch and pre-release identifiers in that order (Build metadata does not figure into precedence). ...

See spec items [10](http://semver.org/#spec-item-10) and [11](http://semver.org/#spec-item-11).

Changelog:

* Remove build metadata comparison in `compare_by_keys`
* Update `test_should_compare_versions_with_build_and_release`
* Add `test_should_ignore_builds_on_compare`

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -57,17 +57,14 @@ def compare(ver1, ver2):
         rc1, rc2 = d1.get('prerelease'), d2.get('prerelease')
         rccmp = nat_cmp(rc1, rc2)
 
-        build_1, build_2 = d1.get('build'), d2.get('build')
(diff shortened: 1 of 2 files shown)


k-bx:
@jeloagnasin nice, thank you!

k-bx:
Released under 2.6.0

jeloagnasin:
@k-bx, no prob! Thank you too for this library. 👍

PR #42, opened 2016-12-19
Improve Usability with Docstrings

tomschr:
I think, it is really important to have docstrings as a kind of good practice and for introspection when working in the interactive Python shell.

Find the pull request with docstrings and a small test case. ;-)

Hope you find it useful.


k-bx:
@tomschr thank you, released under 2.7.3!

tomschr:
@k-bx Thanks! :+1:

PR #43, opened 2016-12-23
Add support for dot in patch part and major.minor version only

minhkiller:
Add support for "." in patch and major.minor version only.
For example: 1.0, 3.4.5.Final, 4.2.5.RELEASE

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,15 +9,17 @@ All notable changes to this code base will be documented in this file,
 in every released version.
 
 
-Version 2.7.0
+Version 2.7.2
 =============
 
-:Released: FUTURE
-:Maintainer: UNKNOWN <unknown>
+:Released: 2016-11-08
+:Maintainer: Kostiantyn Rybnikov <k-bx@k-bx.com>
 
 Additions
 ---------
 
+* Fix issue #37 (Remove trailing zeros from prelease doesn't allow to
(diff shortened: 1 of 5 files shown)


minhkiller:
Hi @k-bx,

I created the PR to support "." in patch and version with major.minor only
For example: 1.0, 3.4.5.Final (hibernate library), 4.2.5.RELEASE (Spring framework)

Could you please help review and consider to merge this upstream?

Thanks,
Minh Tran

k-bx:
@minhkiller thanks for the work, but unfortunately I think I can't merge this. Semver clearly states that PATCH version is an integer, and only allows hyphen to follow it.

I suggest either adding some pre-filtering function which would transform hibernate and sprint-like version into a form which is a valid semver and use this library on top, or to write a complete parser, whatever fits your use-case best.

Thank you for the effort anyways!

minhkiller:
@k-bx Thanks for your response.
Can you accept a lenient mode for the parser? Certainly I'm happy with rejecting shitty version strings by default

k-bx:
@minhkiller actually, yeah, why not. Please make a separate function `parse_lanient` or similar name and make a clear documentation for it, I see no problem having it in-place.

minhkiller:
Cool thanks @k-bx. I will update the PR

PR #46, opened 2017-01-16
fix version comparisons

### LINKED ISSUE #45: Version comparison does not conform to semver spec for prerelease tags.
The semver spec from http://semver.org/ states the following under [section 11](http://semver.org/#spec-item-11) of the spec with respect to how version precedence should be computed.

```
1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0
```

This library fails on some of these comparisons.

```python
In [6]: pairs = (
 ('1.0.0-alpha', '1.0.0-alpha.1'),
 ('1.0.0-alpha.1', '1.0.0-alpha.beta'),
 ('1.0.0-alpha.beta', '1.0.0-beta'),
 ('1.0.0-beta', '1.0.0-beta.2'),
 ('1.0.0-beta.2', '1.0.0-beta.11'),
 ('1.0.0-beta.11', '1.0.0-rc.1'),
 ('1.0.0-rc.1', '1.0.0'))

In [7]: for a, b in pairs:
   ...:     print("Checking {0} < {1} =>".format(a, b), semver.cmp(a, b) == -1)
   ...:
Checking 1.0.0-alpha < 1.0.0-alpha.1 => True
Checking 1.0.0-alpha.1 < 1.0.0-alpha.beta => True
Checking 1.0.0-alpha.beta < 1.0.0-beta => True
Checking 1.0.0-beta < 1.0.0-beta.2 => True
Checking 1.0.0-beta.2 < 1.0.0-beta.11 => False
Checking 1.0.0-beta.11 < 1.0.0-rc.1 => True
Checking 1.0.0-rc.1 < 1.0.0 => False
```

Specifically, `1.0.0-beta.2 < 1.0.0-beta.11` and `1.0.0-rc.1 < 1.0.0` do not resolve to the correct comparison.

pipermerriam:
Fixes #45 

### What was wrong?

The logic implementing the prerelease version comparisons was not quite correct.

### How was it fixed?

Modified how pre-release tags are compared to ensure that all of the example test vectors from the semver spec evaluate correctly.

#### Cute animal picture

baby-kangaroo-white

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -70,10 +70,24 @@ def test_should_parse_zero_prerelease():
 
 def test_should_get_less():
     assert compare("1.0.0", "2.0.0") == -1
+    assert compare('1.0.0-alpha', '1.0.0-alpha.1') == -1
+    assert compare('1.0.0-alpha.1', '1.0.0-alpha.beta') == -1
+    assert compare('1.0.0-alpha.beta', '1.0.0-beta') == -1
+    assert compare('1.0.0-beta', '1.0.0-beta.2') == -1
+    assert compare('1.0.0-beta.2', '1.0.0-beta.11') == -1
+    assert compare('1.0.0-beta.11', '1.0.0-rc.1') == -1
+    assert compare('1.0.0-rc.1', '1.0.0') == -1
 
 
 def test_should_get_greater():
     assert compare("2.0.0", "1.0.0") == 1
+    assert compare('1.0.0-alpha.1', '1.0.0-alpha') == 1
+    assert compare('1.0.0-alpha.beta', '1.0.0-alpha.1') == 1
+    assert compare('1.0.0-beta', '1.0.0-alpha.beta') == 1
+    assert compare('1.0.0-beta.2', '1.0.0-beta') == 1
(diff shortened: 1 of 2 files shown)


pipermerriam:
Let me know if there is anything about this implementation you would like changed.

tests.py
```diff
@@ -232,7 +246,7 @@ def test_prerelease_order():
     assert min_ver('1.2.3-Rc10', '1.2.3-rc10') == '1.2.3-Rc10'
     # Numeric identifiers always have lower precedence than non-numeric
     # identifiers.
-    assert min_ver('1.2.3-2', '1.2.3-rc') == '1.2.3-rc'
+    assert min_ver('1.2.3-2', '1.2.3-rc') == '1.2.3-2'
```
k-bx: nice catch!

k-bx:
@pipermerriam released under 2.7.4, thank you!

Issue #52, opened 2017-03-30
Complex filters


kpcyrd:
hi, I'm trying to use python-semver for more complex comparison :)

```
>>> semver.match('6.2.3', '>=6,<6.2.1|>=4.0.0-rc2,<4.2.4|>=5,<5.3.1')
```
which should match according to the rules [composer][1] does it. This is obviously currently not supported but I would really love it to have a method for something like this :)

[1]: https://getcomposer.org/

k-bx:
Hi! You can use a simple Python operators meanwhile to combine comparisons, but I wouldn't mind having a separate API for parsing something like this.

Care to make a PR?

kpcyrd:
@k-bx I've hacked together a bunch of code to parse it successfully, I'm trying to clean it up.

Since you're the author of an semver parser, how should I handle `>=6`?

This currently causes an exception because it's not valid semver and I currently translate it to `>=6.0.0`, which I'm not sure is correct. Do you have an opinion on this?

k-bx:
Yeah. My opinion is — need to make sure semver code is strictly following semver spec, and having the rest additional functionality kept separate and clearly stating that :)

scls19fr:
@kpcyrd may you should ask at https://github.com/semver/semver to improve semver specification to add this concept of complex comparison.
When a "standard" will emerge we will happily accept PR about it.

PR #53, opened 2017-03-31
(not semver) Support complex filters

kpcyrd:
Implements #52.

This currently doesn't cover my usecase as I'd have to pass a custom function to pre-process non-semver complient versions.

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -195,6 +195,38 @@ def match(version, match_expr):
     return cmp_res in possibilities
 
 
+def match_all(version, match_exprs):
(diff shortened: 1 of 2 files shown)


tomschr:
Thanks @kpcyrd your contribution is much appreciated although it's a long time ago, unfortunately.

I would like to see this in this extended version comparison issue (linked above). Probably as a class. If you like, you can look into this.

tlaferriere:
I would say this PR is stale. Maybe we should close it @tomschr ?

Issue #54, opened 2017-05-11
Object comparison works different than function comparison


Chive:
Hi there!

I'm trying to use object comparison but I'm not getting the expected results:

```python
>>> import semver

>>> v1 = '0.10.4'
>>> v2 = '0.10.4-beta.1'

"""
from: semver#compare
The return value is negative if ver1 < ver2,
zero if ver1 == ver2 and strictly positive if ver1 > ver2
"""
>>> semver.compare(v1, v2) > 0
True
>>> semver.compare(v2, v1) < 0
True

# until here, everythings works as expected

>>> v1o = semver.parse_version_info(v1)
>>> v2o = semver.parse_version_info(v2)

>>> v1o
VersionInfo(major=0, minor=10, patch=4, prerelease=None, build=None)
>>> v2o
VersionInfo(major=0, minor=10, patch=4, prerelease='beta.1', build=None)

# this is where things go wrong:
>>> v1o > v2o
False
>>> v2o < v1o
False

```

k-bx:
Fixed by #55

PR #55, opened 2017-05-19
Add comparision between VersionInfo objects

ppkt:
I updated `VersionInfo` class to inherit from named tuple and I added rich comparison methods to allow compare between two `VersionInfo` instances or between `VersionInfo` and dictionary.
This PR should resolve issue #54.

Since we are still supporting Python 2.6 it was not possible to use [functools.total_ordering](https://docs.python.org/3.6/library/functools.html#functools.total_ordering).

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -4,7 +4,6 @@ Python helper for Semantic Versioning (http://semver.org/)
 
 import collections
 import re
-import sys
 
 
 __version__ = '2.7.6'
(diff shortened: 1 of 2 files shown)


k-bx:
Thank you! I'm not sure I understand the error for python2.6, will check it out and merge then. PR looks good

ppkt:
For me it looks like problem with Python interpreter:  [ERROR: InterpreterNotFound: python2.6
](https://travis-ci.org/k-bx/python-semver/jobs/234168946#L194)

Unfortunately I also wasn't able to install Py2.6 on my local machine to test my changes (using pyenv).

k-bx:
So, tests fail on master same way, probably travis updated their image and now it doesn't have 2.6 interpreter there. Anyways, I'll create a separate ticket to handle that, merging meanwhile.

k-bx:
Released under 2.7.7

Issue #57, opened 2017-05-25
Fix travis under 2.6


k-bx:
For some reason, that interpreter stopped being installed in travis

ppkt:
@k-bx I just checked Travis documentation and it seems that there might be a problem with tox and Travis. It's because we are asking Travis to configure environment only for Python 3.5 (see [here](https://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against)).

Maybe external library, like [tox-travis](https://github.com/ryanhiebert/tox-travis) could help?

k-bx:
@ppkt yeah, maybe. Not sure I'll have time for this in the nearest time, so PRs are welcome!

PR #58, opened 2017-05-26
Update Travis configuration file

ppkt:
I updated a Travis config file to use Tox with tested Python version. It should fix problems with missing interpreter (see output [here](https://travis-ci.org/ppkt/python-semver)) and resolve issue #57.

I also noticed that there is a problem with Python 3.2 - I'll take a look on this problem in free time.


k-bx:
A bit thank you!

PR #62, opened 2017-08-25
support custom default names for pre and build

ofek:
e.g. `dev` instead of `rc` or `nightly` instead of `build`


ofek:
@k-bx Hi there! How does this look?

k-bx:
@ofek thanks! Released under 2.7.8

ofek:
Thank you!

PR #64, opened 2017-08-27
restructure project

ofek:
adheres a bit more to standards

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,4 +13,4 @@ setenv =
 [testenv:flake8]
 basepython = python3.4
 deps = flake8
-commands = flake8 . --exclude="__init__.py"
+commands = flake8
--- a/tox.ini
+++ b/tox.ini
@@ -13,4 +13,4 @@ setenv =
 [testenv:flake8]
 basepython = python3.4
 deps = flake8
-commands = flake8 --ignore=F401
+commands = flake8 --exclude=__init__.py
--- a/tox.ini
+++ b/tox.ini
@@ -13,4 +13,4 @@ setenv =
 [testenv:flake8]
 basepython = python3.4
 deps = flake8
-commands = flake8
+commands = flake8 --exclude="__init__.py"
--- a/tox.ini
+++ b/tox.ini
@@ -13,4 +13,4 @@ setenv =
 [testenv:flake8]
 basepython = python3.4
 deps = flake8
-commands = flake8 --exclude="__init__.py"
+commands = flake8 semver --exclude="__init__.py"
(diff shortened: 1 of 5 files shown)


ofek:
@k-bx Hello again! Please let me know what you think about this. It's basically just a bit more organized now.

k-bx:
@ofek I don't mind restructuring the code to be closer to standards, but may I ask which standards do you mean here? Thanks

tox.ini
```diff
@@ -5,12 +5,12 @@ envlist =
...
+commands = py.test
 deps = pytest
 setenv =
     PIP_DISABLE_PIP_VERSION_CHECK = 1
 
 [testenv:flake8]
 basepython = python3.4
 deps = flake8
-commands = flake8
+commands = flake8 --ignore=F401
```
k-bx: Why was skipping the F401 was necessary?

ofek: Since __init__.py imports what is needed into the root namespace, but doesn't use anything itself, it detects unused variables. It seems other projects do this too:

https://github.com/requests/requests/blob/5dcbdb50fc9652cbeffeec22a21d1196e938bfca/Makefile#L14

ofek:
http://docs.python-guide.org/en/latest/writing/structure

Also, users can just `git clone https://github.com/k-bx/python-semver && cd python-semver` then simply `pytest` which is what most people expect to work.

Your coverage is quite good btw, 97%!

k-bx:
@ofek sorry, I just don't see how this makes things any better. The standard way of running tests is by running `python setup.py test`, it's also mentioned in the README. I don't think that making default `py.test` work would be a big enough deal to convert simple one-file library in a whole package. I'm fine with adding `py.test -q tests.py` in README if someone is not able to figure to look in tox.ini, though I'm not too worry since Travis will run tests for them in a PR anyway.

Arguing that other packages are doing something bad (removing flake checks) is rather the reason to fix those packages, not make this one worse. I believe you are able to remove a check on a per-module basis, e.g. removing the check from `__init__.py` instead of removing it from the whole package. Removing it from the whole package would remove the check even from those modules which could benefit from such a check.

ofek:
@k-bx Thanks! Just fixed that.

ofek:
Hmm, dunno what happened here. I'll look into it.

ofek:
@k-bx Ok, flake8 is actually fixed now.

scls19fr:
Thanks @ofek for your contribution.
However, I think this PR adresses too much issues at once.
So I'm closing. Feel free to create an issue for each "problem" you want to address and submit after a PR.

PR #71, opened 2018-01-02
(not semver) Obj, Non-Strict, default 0, 

prikkeldraad:
I tried to rewrite semver to class with backwards compatibility with more python-like methods: format, cmp, max and min.

I also added a default for minor and patch in case those are not supplied. In case version 1 is compared to 1.0 or 1.0.0 they should all match equal.

Last thing is non-strict testing which compares also the buildnumber. Although this is not semver specific, when sorting builds, there is a need to know what the latest one is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 /MANIFEST
 
 *.py[co]
+*.swp
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,12 @@ Python SemVer library
 All notable changes to this code base will be documented in this file,
 in every released version.
 
+Version 2.8.0
+=============
+* Backwards compatible change to use objects
+* Versions will default to 0 for minor and patch when not supplied
+* Semver object will take string, tuple or list for version
+
 Version 2.7.9
 =============
 
--- a/README.rst
+++ b/README.rst
@@ -69,13 +69,17 @@ This module provides just couple of functions, main of which are:
     '1.0.0'
 
 Old style
+
 .. code-block:: python
+
    >>> import semver
    >>> semver.parse('4.0')
    4.0.0
 
 New style
-.. code-block:: python
+
+.. code-block:: pythona
+
    >>> from semver import semver
    >>> version = semver('3.2.1')
    >>> print version
--- a/README.rst
+++ b/README.rst
@@ -78,12 +78,21 @@ Old style
 
 New style
 
-.. code-block:: pythona
+.. code-block:: python
 
    >>> from semver import semver
    >>> version = semver('3.2.1')
    >>> print version
    3.2.1
+
+Non-strict comparison
+
+.. code-block:: python
+
+   >>> from semver import semver
+   >>> version = semver('1.2.3-alpha+1234')
+   >>> print version.cmp(semver('1.2.3-alpha+1235'), strict=False)
+   1 
    
 Installation
 ------------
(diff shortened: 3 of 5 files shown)


kxepal:
- There is no need in class at all, especially in the one which is badly designed in the name of backward compatibility. Why there should be the one?

- Optional minor and patch versions is the change that should be discussed separately. Is that even legal? Semver spec doesn't defines them as optionals.

- Why strict param and how it relates to the spec? Should this project take the way of creating own vision of what semver is?

- What is swp file to ignore?

prikkeldraad:
> There is no need in class at all, especially in the one which is badly designed in the name of backward compatibility. Why there should be the one?

I think the class approach looks better and cleaner. Instead of semver.parse('1.2.3') use semver('1.2.3'), with this approach chaining is also possible. I'm aware that my skills might be a bit outdated and refactoring to a better design might be in place

> Optional minor and patch versions is the change that should be discussed separately. Is that even legal? Semver spec doesn't defines them as optionals.

Semver states that these should be there and are mandatory, however in an automated environment where not all applications follow semver, defaulting to zero on minor and patch eases automation of ALM. This solution doesn't break semvers implementation or rules, it add's functionality to use in an automated ALM environment

> Why strict param and how it relates to the spec? Should this project take the way of creating own vision of what semver is?

Semver does not take in account build numbers in comparison. In ALM this might be important. To not break semver I introduced the strict option. That option is default True and uses the semver ruleset. When buildnumber diffence is needed however we need to have a way to compare those. Only when strict is False (ergo not following semver rules complete) we can compare with buildnumbers.

> What is swp file to ignore?

Committing when a file is in edit mode in vim it will add .swp (the file which is being edited). So .swp should be added in .gitignore

kxepal:
>> There is no need in class at all, especially in the one which is badly designed in the name of backward compatibility. Why there should be the one?
>
>I think the class approach looks better and cleaner. Instead of semver.parse('1.2.3') use semver('1.2.3'), with this approach chaining is also possible. I'm aware that my skills might be a bit outdated and refactoring to a better design might be in place

Quite doubtful profit, imho. Current approach: there is plain data (string), plain function and plain result (dict) - quite trivial. Proposed: there is custom object that encapsulates some data and provides set of properties/methods to work with - one more abstraction to learn.

>> Optional minor and patch versions is the change that should be discussed separately. Is that even legal? Semver spec doesn't defines them as optionals.
>
>Semver states that these should be there and are mandatory, however in an automated environment where not all applications follow semver, defaulting to zero on minor and patch eases automation of ALM. This solution doesn't break semvers implementation or rules, it add's functionality to use in an automated ALM environment

If they don't follow semver, why semver rules should be applied to them? For instance, application uses calendar strategy (ala Ubuntu) - proposed changes will allow semver module to work with such, but does this means that semver will do it right?

>> Why strict param and how it relates to the spec? Should this project take the way of creating own vision of what semver is?
>
> Semver does not take in account build numbers in comparison. In ALM this might be important.

What do you mean? Here is the test about build comparison:
https://github.com/k-bx/python-semver/blob/master/tests.py#L160
Also, what is ALM you reference?

tomschr:
This is quite an old PR and a lot of things have changed now. There are even conflicts with some files. For the time being, I close this issue as some parts are already integrated.

If you still feel the need, just open a new PR and we can discuss it there.

Thanks for your contribution and hope you understand this.

Issue #74, opened 2018-05-11
Measure and improve code coverage


scls19fr:
Hello,

You should measure [code coverage](https://en.wikipedia.org/wiki/Code_coverage).
`parse_version_info`, for example, doesn't have test(s).

Kind regards

tomschr:
> `parse_version_info`, for example, doesn't have test(s).

Actually it does: it contains doctests inside the docstring. These also count as tests. :wink:

Issue #73, opened 2018-05-11
Make VersionInfo Hashable


scls19fr:
Hello,

`VersionInfo` doesn't implement `__hash__`

```python
import click
import datetime
import requests
import requests_cache
import pandas as pd
from pandas.io.json import json_normalize
pd.set_option("max_rows", 10)
import semver

@click.command()
@click.option('--cache', default=7, help='Number of days to cache queries.')
def main(cache):
    expire_after = datetime.timedelta(days=cache)
    session = requests_cache.CachedSession(cache_name='cache', backend='sqlite', expire_after=expire_after)
    url = "http://downloads.arduino.cc/libraries/library_index.json"
    r = session.get(url)
    d = r.json()
    libraries = d['libraries']
    df = json_normalize(libraries)
    df['version'] = df['version'].map(semver.parse_version_info)  # TypeError: unhashable type: 'VersionInfo
    df.sort_values(by=['name', 'version'], ascending=[True, False])
    #print(df)
    #df.to_excel("libraries.xlsx")


if __name__ == '__main__':
    main()
```

raises `TypeError: unhashable type: 'VersionInfo'`

This is quite close to #72 

Kind regards

scls19fr:
An other approach for hashing `VersionInfo` (differently than hashing it's string representation) could be to get a dict from `VersionInfo` but dict are not hashable (only immutable objects are hashable according https://stackoverflow.com/questions/13264511/typeerror-unhashable-type-dict).
So we can convert dict to frozenset using

```python
def freeze(d):
    if isinstance(d, dict):
        return frozenset((key, freeze(value)) for key, value in d.items())
    elif isinstance(d, list):
        return tuple(freeze(value) for value in d)
    return d
```

and called `__hash__` of this `frozenset`

Here is an example:

```python
In [51]: v = semver.parse_version_info('1.2.3')

In [52]: d = v._asdict()

In [53]: fs = freeze(d)

In [54]: fs.__hash__()
Out[54]: -1214760729421905988
```

scls19fr:
A last approach (and probably sympler) could be to get a tuple from a VersionInfo

    def __hash__(self):
        return hash(self._astuple())

PR #75, opened 2018-05-11
Implements __str__ and __hash__

### LINKED ISSUE #72: VersionInfo string representation
It seems that the string representation is not available from `VersionInfo`. It should be quite straightforward to do with a custom string representation method:

```python
def __str__(self) -> str:
    return (str(version.major) or '') \
           + (('.' + str(version.minor)) if version.minor is not None else '') \
           + (('.' + str(version.patch)) if version.patch is not None else '') \
           + (('-' + str(version.prerelease)) if version.prerelease is not None else '') \
           + (('+' + str(version.build)) if version.build is not None else '')
```

Note that we can keep `__repr__` untouched so that it provides unambiguous representation.

scls19fr:
Closes #72, #73
Should help #74

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -106,6 +106,17 @@ class VersionInfo(collections.namedtuple(
             return NotImplemented
         return _compare_by_keys(self._asdict(), _to_dict(other)) >= 0
 
+    def __str__(self):
+        return (str(self.major) or '') \
+            + (('.' + str(self.minor)) if self.minor is not None else '') \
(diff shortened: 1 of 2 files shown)


scls19fr:
Thanks to @smarie for a part of this code

k-bx:
@scls19fr hey, would you like to become a co-maintainer of the package? I don't think I have time maintaining the low-priority things like this one. If yes, please send me your PYPI username as well.

PR #80, opened 2018-05-15
Remove Python 2.6 support

### LINKED ISSUE #76: Is compatibility with Python 2.6 version still required?
Hello,

I'm just asking this because of
https://travis-ci.org/k-bx/python-semver/builds/377894321

but that's a bad reason :smiley:

Kind regards

scls19fr:
Closes #76

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ install: pip install "virtualenv<14.0.0"
 script: python setup.py test
 matrix:
   include:
-  - python: "2.6"
-    env: TOXENV=py26
-
   - python: "2.7"
     env: TOXENV=py27
 
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,6 @@ setup(
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.2',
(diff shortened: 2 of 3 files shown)


scls19fr:
@k-bx can we merge this ?
Other PR can't be merged safely without this one

ppkt:
@scls19fr @k-bx I guess in this case we could also drop support for Python 3.2 and 3.3:
https://www.python.org/downloads/release/python-326/
https://www.python.org/downloads/release/python-337/

scls19fr:
For now CI for Python 2.6 is failing and prevent us from merging safely others PR.
This is (currently) not the case for Python 3.2 and 3.3 so that's the reason why I'm just considering for now removing 2.6.
Removing other Python versions can probably be asked in a new issue and adressed in an other python-semver release.

ppkt:
@scls19fr I see, but since you're going to release version 2.8.0 it might be a good moment to remove other versions too.

kxepal:
Btw, removing support of those is breaking change or not? Eventually or accidentally, semver could break compatibility with them because CI won't make any checks against. Feels like good idea to drop obsolete Python versions support with a major release.

scls19fr:
I'd prefer to avoid breaking too much eggs for a first omelette.
Maybe removing Python 3.2 and 3.3 should be considered when releasing python-semver 2.9.0
Anyway, feel free to open issue to keep a track of this discussion about Python 3.2, 3.3 removal.

k-bx:
My IMHO is that it's somewhat embarrassing to not support as many pythons as possible for something as simple as a semver-parsing, but I don't have any interest in supporting 2.6, so I don't have strong feelings dropping it. Do whatever you feel like.

scls19fr:
Good bye Python 2.6

PR #93, opened 2018-05-26
Remove namedtuple inheritance

### LINKED ISSUE #87: Bad rendering of VersionInfo in Pandas DataFrame
Hello,

I noticed that `VersionInfo` are rendered badly in Pandas DataFrame.
(see https://github.com/scls19fr/arduino_libraries_search/issues/8 )

They are rendered as tuple.

```python
In [1]: import pandas as pd
In [2]: import semver
In [3]: versions = ['4.0.0', '5.0.0', '5.0.1', '10.0.0']
In [4]: df = pd.DataFrame(versions, columns=['versions'])
In [5]: df  # version as string
In [6]: df.sort_values(by='versions')  # so they are not ordered correctly
Out[6]:
  versions
3   10.0.0
0    4.0.0
1    5.0.0
2    5.0.1
In [7]: df['versions'] = df['versions'].map(semver.parse_version_info)  # parse versions as `VersionInfo`
In [8]: df.sort_values(by='versions')  # VersionInfo are ordered correctly but are badly rendered
Out[8]:
                 versions
0   (4, 0, 0, None, None)
1   (5, 0, 0, None, None)
2   (5, 0, 1, None, None)
3  (10, 0, 0, None, None)
```

I don't know what method Pandas use for rendering them.
What I see is that it's neither using `__str__` neither using `__repr__` method

```python
In [9]: v = df['versions'].iloc[-1]

In [10]: v
Out[10]: VersionInfo(major=10, minor=0, patch=0, prerelease=None, build=None)

In [11]: v.__str__()
Out[11]: '10.0.0'

In [12]: v.__repr__()
Out[12]: 'VersionInfo(major=10, minor=0, patch=0, prerelease=None, build=None)'
```

`VersionInfo` inherits from `namedtuple` so this is probably related.

Any idea?

### LINKED ISSUE #94: VersionInfo shouldn't inherit from namedtuple
It should help fixing #87

scls19fr:
Closes #87 
Closes #94

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -73,8 +73,7 @@ def parse(version):
     return version_parts
 
 
-class VersionInfo(collections.namedtuple(
-        'VersionInfo', 'major minor patch prerelease build')):
+class VersionInfo:
     """
     :param int major: version when you make incompatible API changes.
     :param int minor: version when you add functionality in
@@ -83,7 +82,26 @@ class VersionInfo(collections.namedtuple(
     :param str prerelease: an optional prerelease string
     :param str build: an optional build string
     """
-    __slots__ = ()
+
+    def __init__(self, major, minor, patch, prerelease, build):
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+        self.prerelease = prerelease
+        self.build = build
+
+    def _astuple(self):
+        return (self.major, self.minor, self.patch,
+                self.prerelease, self.build)
+
+    def _asdict(self):
+        return collections.OrderedDict((
+            ("major", self.major),
+            ("minor", self.minor),
+            ("patch", self.patch),
+            ("prerelease", self.prerelease),
+            ("build", self.build)
+        ))
 
     def __eq__(self, other):
         if not isinstance(other, (VersionInfo, dict)):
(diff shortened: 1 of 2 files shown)


semver.py
```diff
@@ -83,7 +82,26 @@ class VersionInfo(collections.namedtuple(
     :param str prerelease: an optional prerelease string
     :param str build: an optional build string
     """
-    __slots__ = ()
```
kxepal: So now VersionInfo is a mutable object with dynamic attrs. That's quite too dramatic change in the name of pretty printting pandas dataframes /:

scls19fr: Nothing else than Pandas printing but that's quite problematic as it won't be solved soon (my 2 cts)

Some tests to ensure that VersionInfo are immutable should be added if mutability of VersionInfo is a problem

hugobuddel:
It seems this change broke backwards compatibility, because VersionInfo is not iterable anymore.

See https://github.com/maartenbreddels/releash/pull/2 or https://travis-ci.org/maartenbreddels/releash/jobs/439551331 , which worked 7 months ago with semver 2.7.9, but fails now with 2.8.1.

Ironic this happened in the `semver` package. Should this have been merged into 3.0.0, and not in 2.x.y?

hugobuddel:
Thanks @ppkt! It seems okay to have this change in a minor version increase because it was supposed to be backwards compatible. The old behavior could be considered undocumented and/or the new behavior can be seen as a bug and be fixed on a patch level (which you just did).

Thanks for fixing it so quickly. I've just written out the list, so that code works with all versions of semver.

PR #97, opened 2018-05-30
Immutable VersionInfo

### LINKED ISSUE #96: VersionInfo shouldn't be mutable
From https://github.com/k-bx/python-semver/pull/93#discussion_r191442817

We should add tests to ensure immutability of VersionInfo

scls19fr:
Closes #96

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -73,7 +73,7 @@ def parse(version):
     return version_parts
 
 
-class VersionInfo:
+class VersionInfo(object):
     """
     :param int major: version when you make incompatible API changes.
     :param int minor: version when you add functionality in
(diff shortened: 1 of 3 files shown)


scls19fr:
I don't understand why CI is broken with pypy and Python 2.7.

Any idea?

semver.py
```diff
@@ -82,13 +82,40 @@ class VersionInfo:
...
+
+    @property
+    def prerelease(self):
+        return self._prerelease
+
+    @property
+    def build(self):
+        return self._build
+
+    def __setattr__(self, name, value):
```
ppkt: I think this function is unnecessary

test_semver.py
```diff
@@ -414,3 +414,42 @@ def test_parse_method_for_version_info():
     s_version = "1.2.3-alpha.1.2+build.11.e0f985a"
     v = VersionInfo.parse(s_version)
     assert str(v) == s_version
+
+
+def test_immutable():
```
ppkt: Could you also add a test to check if it's possible to add new attribute? Something like this:
```
try:
    v.some_attribute = 1
except:
    ...
```

semver.py
scls19fr: I thought also that __setattr__ wasn't necessary...
but not with Python 2.7 and Pypy

kxepal: Why do we want to have private attributes be mutable?

semver.py
```diff
@@ -82,13 +82,40 @@ class VersionInfo:
...
+        if name not in self.__slots__:
+            raise AttributeError
```
kxepal: Even if so, it's better to pass name to AttributeError to avoid any confusions.

semver.py
scls19fr: You are right ... 
[This build](https://travis-ci.org/k-bx/python-semver/builds/386245547?utm_source=github_status&utm_medium=notification) shows that  `__setattr__` is not required.

Although what is required for Python 2.7 and Pypi is inheriting VersionInfo from `object`.

PR #104, opened 2018-07-17
Fix comparison

### LINKED ISSUE #102: Replicating usage docs results in '<' not supported between instances of 'VersionInfo' and 'tuple'
I feel like I must be missing something, but when I follow the docs [here](http://python-semver.readthedocs.io/en/latest/usage.html), I get:

```
In [1]: import semver

In [2]: semver.__version__
Out[2]: '2.8.1'

In [3]: from semver import VersionInfo

In [4]: version_info = VersionInfo.parse("3.4.5-pre.2+build.4")

In [5]: version_info
Out[5]: VersionInfo(major=3, minor=4, patch=5, prerelease='pre.2', build='build.4')

In [6]: version_info > (1, 0)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-d89d50dc6fc1> in <module>()
----> 1 version_info > (1, 0)

TypeError: '>' not supported between instances of 'VersionInfo' and 'tuple'

In [7]: version_info < (3, 5)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-3076f0f266b0> in <module>()
----> 1 version_info < (3, 5)

TypeError: '<' not supported between instances of 'VersionInfo' and 'tuple'
```

### LINKED ISSUE #103: Wrong comparison result when comparing VersionInfo and string
See https://github.com/k-bx/python-semver/issues/102#issuecomment-405467250

scls19fr:
Closes #102
Closes #103

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -4,9 +4,12 @@ Python helper for Semantic Versioning (http://semver.org/)
 
 import collections
 import re
+import sys
+
+PY2 = sys.version_info[0] == 2
 
 
-__version__ = '2.8.1'
+__version__ = '2.8.2'
 __author__ = 'Kostiantyn Rybnikov'
 __author_email__ = 'k-bx@k-bx.com'
 __maintainer__ = 'Sebastien Celles'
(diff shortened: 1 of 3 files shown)


semver.py
```diff
@@ -261,6 +257,13 @@ def _compare_by_keys(d1, d2):
     return rccmp
 
 
+def _ensure_is_comparable(other):
+    comparable_types = (VersionInfo, dict, tuple)
+    if not isinstance(other, comparable_types):
+        raise NotImplementedError("other type %r must be in %r"
```
kxepal: https://docs.python.org/3/library/constants.html#NotImplemented

You need to return NotImplemented here. Python itself will raise proper TypeError if objects aren't comparable.

semver.py
```diff
@@ -261,6 +257,13 @@ def _compare_by_keys(d1, d2):
     return rccmp
 
 
+def _ensure_is_comparable(other):
+    comparable_types = (VersionInfo, dict, tuple)
```
kxepal: @ppkt made a good point about using interfaces in place of dict/tuple. This will allow us to compare with dict-like / tuple-like objects for free (well, not really, but nothing to care about).

The only thing is `str` - it's Iterable, but unlikely `*str` will give us what we really expected from it.

semver.py
scls19fr: Thanks @kxepal 
I also noticed https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror

semver.py
```diff
@@ -125,33 +128,51 @@ def _asdict(self):
...
+            else:
+                return NotImplemented
         return _compare_by_keys(self._asdict(), _to_dict(other)) > 0
 
     def __ge__(self, other):
-        if not isinstance(other, (VersionInfo, dict)):
-            return NotImplemented
+        if not _is_comparable(other):
+            if PY2:
+                raise TypeError
```
kxepal: Hm, but why? NotImplemented existed for Python 2.x as well and there rules are the same for it.

scls19fr: See previous build https://travis-ci.org/k-bx/python-semver/builds/404887965?utm_source=github_status&utm_medium=notification for Python 2.7 ([see job](https://travis-ci.org/k-bx/python-semver/jobs/404887966))

even if NotImplemented existed for Python 2.x returning it doesn't raise TypeError (contrary to Python 3.x) and comparison wasn't returning a correct result.

ppkt: No, NotImplemented in Py2.7 works in the same way: https://docs.python.org/2.7/library/constants.html#NotImplemented

kxepal: Ah, Python 2.x swallows them and returns always True or False...or False in our case.

kxepal: I think, it would be better to move that check and behaviour difference into tests. 

While it's good idea to keep same behaviour in both Pythons, Python 2.x world tells that every comparison end with boolean result, so we would looks here like a rule breaker.

Raising TypeError for Python 2.x also breaks NotImplemented behavior when if `a.__eq__(b)` fails, Python tries to do `b.__eq__(a)` - by raising exception we drops the second and our behaviour in both Pythons still remains inconsistent.

Issue #70, opened 2017-12-20
(not semver) Support for less than major.minor.patch


prikkeldraad:
Althoug semver states that m.m.p should be present. That does not mean it always is. Some software only gives a major or major.minor number. I would like semver to parse those numbers with minor or patch numbers defaulting to 0. That's what is ment by those major only numbers anyway.

`version = semver.parse('3.0')`
Should result in 3.0.0

`version = semver.parse('10')`
Should result in 10.0.0

k-bx:
I wouldn't mind having something like this, but it should be under an explicit API, separate from `semver.parse`, that would behave in a more "relaxed" way, documenting the differences in detail

scls19fr:
Can we consider that passing a parameter named `strict` (with `True` as default) can be a solution ?

To be able to parse this kind of string we will have to do:

    version = semver.parse('10', strict=False)

I wonder if semver spec shouldn't be improved (asking at https://github.com/semver/semver ) to be more explicit about how parsing a version should behave

k-bx:
No, please make a separate function with an obvious name that it's not Semver but something different. parse_relaxed might do

scls19fr:
Ok

Adding a function such as `issemver` could also be interesting.

Maybe we should implement it this way.
```python
try:
    semver.parse(version)
    return True
except Exception as e:
    return False
```

kxepal:
```
try:
    semver.parse(version)
except ValueError:
    return False
else:
    return True
```
If so.

tomschr:
This issue is similar to others that we've closed. :wink: 

Currently, this is out of scope. However, we've documented it how to deal with such cases. Have a look at https://python-semver.readthedocs.io/en/latest/usage.html#dealing-with-invalid-versions

Thanks for all the input.

PR #108, opened 2018-10-07
(not semver) Allow parsing of partial version string

spack971:
Allow `semver` to parse partial version string. 

When only a single digit is given, it is assumed to be the major version: `1` is equivalent to `1.0.0`.

When two digits are given, it is assumed to be the major and minor parts of the version string: `1.2` is equivalent to `1.2.0`.

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -31,6 +31,18 @@ _REGEX = re.compile(
         $
         """, re.VERBOSE)
 
+_REGEX_PARTIAL = re.compile(
+    r"""
+    ^
+    (?P<major>(?:0|[1-9][0-9]*))
+    (?:
+        \.(?P<minor>(?:0|[1-9][0-9]*))
+    )?
+    $
+    """,
+    re.VERBOSE
+)
+
 _LAST_NUMBER = re.compile(r'(?:[^\d]*(\d+)[^\d]*)+')
 
 if not hasattr(__builtins__, 'cmp'):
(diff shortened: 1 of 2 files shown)


ppkt:
@spack971 Thank you for your PR. While it looks interesting, this feature is not compatible with semver: https://semver.org/#spec-item-2. Please check discussion under issue #70.

spack971:
@ppkt Fair point. I'll adapt the pull request to define a `parse_partial` function for this purpose.

spack971:
Added a new function named `try_parse` which will try to parse a complete or partial version string.

scls19fr:
I will personally name such a function `parse_...` because it will be easier to find it when using tab completion with IPython for example

spack971:
@scls19fr well that was my goal but I couldn't come with a nice name. I'm open to suggestions.

tomschr:
That issue looks very similar to #210 which I've done lately. We've come to the conclusion that we do not want to support a function which doesn't adhere to the semver specification. That means, we avoid partial parsing of a version string.

However, to support our developers, we've think to add some recommendations to our documentation is a good idea. As such, I've written a short section into our documentation (see PR #215: "Dealing with Invalid Versions").

@scls19fr With that background, I think, we can close this issue, right?

tomschr:
Closing it as wontfix. There is already a section in the documentation about dealing with invalid versions.

PR #111, opened 2018-10-10
Drop py33

### LINKED ISSUE #110: Travis failing on Python 3.3
I noticed that tox / pytest is causing problems on CI, maybe we should disable running tests for Python 3.3 (or pin to <3.3)?
https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23

scls19fr:
Closes #110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ matrix:
   - python: "2.7"
     env: TOXENV=py27
 
-  - python: "3.3"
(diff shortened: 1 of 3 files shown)


ppkt:
@scls19fr looks nice, thanks - could you add Python 3.7 by the way?

scls19fr:
Yes but I will do in an other PR

PR #120, opened 2019-03-17
Improve test_immutable function

tomschr:

* Raise `AttributeError` with useful messages for major, minor, patch, prerelease, and build attributes.
* Use [`pytest.raises` ](https://docs.pytest.org/en/latest/reference.html?highlight=pytest.raises#pytest-raises) which replace and shorten the `try..except..else` blocks.
* Use `match` keyword in pytest.raises which matches the `AttributeError` from the property setter.
* Split `test_immutable` into several different test functions,  e.g.` test_immutable_major`.

Some comments:

* I've splitted the `test_immutable` function into several other functions as IMHO it makes it more readable.
* If you don't like the property setter and prefer a more cleaner class, I can remove it. I merely added them to have a more correct error message raised which can be tested against.
* The `test_immutable` function could also be parametrized by the `@pytest.mark.parametrize` decorator. However, I've decided against it as the attributes to test are manageable and probably won't change.
* The `test_immutable_unknown_attribute` function just raises the `AttributeError` exception, no error message is checked (like with `test_immutable_major`). I left that out and only checked against the exception.

Hope you like it. :wink:

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -94,22 +94,37 @@ class VersionInfo(object):
     @property
     def major(self):
         return self._major
+    @major.setter
+    def major(self, value):
+        raise AttributeError("attribute 'major' is readonly")
 
     @property
     def minor(self):
         return self._minor
+    @minor.setter
+    def minor(self, value):
+        raise AttributeError("attribute 'minor' is readonly")
 
     @property
     def patch(self):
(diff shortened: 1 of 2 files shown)


tomschr:
Thanks! :+1:

PR #122, opened 2019-03-24
Use python3 instead of python3.4

### LINKED ISSUE #121: tox.ini: should we really specify Python 3.4 for flake8?
## Problem

Let's assume, you have Python 3.6 installed. When using the target `flake8` to check for code style issues, you get the following error:

```
$ tox -e flake8
GLOB sdist-make:.../python-semver/setup.py
flake8 recreate: .../python-semver/.tox/flake8
ERROR: InterpreterNotFound: python3.4
___________________________ summary ____________________________
ERROR:   flake8: InterpreterNotFound: python3.4
```

## Proposed Solution

I would suggest the following minor diff:

```diff
diff --git i/tox.ini w/tox.ini
index 15773af..20024a5 100644
--- i/tox.ini
+++ w/tox.ini
@@ -11,6 +11,6 @@ setenv =
     PIP_DISABLE_PIP_VERSION_CHECK = 1

 [testenv:flake8]
-basepython = python3.4
+basepython = python3
 deps = flake8
 commands = flake8
```

I think, there is no special reason to force users to have Python 3.4 installed on their system. *Any* Python version >= 3.4 can be used to run `flake8`.

Benefits:

* Users are not forced to have Python version 3.4 installed.
* It works for every Python version (as long it is 3.x).

tomschr:
Fix #121. Reduce command `python3.3` to `python3`.

With that patch, it works on Travis and locally. I consider the latter as essential as it doesn't force the user to have Python 3.4 installed. Every 3.x version is possible as long there is an executable or link to `python3`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,6 @@ setenv =
     PIP_DISABLE_PIP_VERSION_CHECK = 1
 
 [testenv:flake8]
-basepython = python3.4
+basepython = python3
 deps = flake8
 commands = flake8


scls19fr:
Thanks @tomschr for your contribution

PR #123, opened 2019-03-24
Improve __repr__ and derive class name from type

tomschr:
Avoids hard coded class name

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -182,7 +182,7 @@ class VersionInfo(object):
     def __repr__(self):
         s = ", ".join("%s=%r" % (key, val)
                       for key, val in self._asdict().items())
-        return "VersionInfo(%s)" % s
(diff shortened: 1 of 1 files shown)


scls19fr:
Maybe adding a test should be considered?

tomschr:
@scls19fr There are already tests, the doctests, for example:

```python
@staticmethod
    def parse(version):
        """Parse version string to a VersionInfo instance.

        >>> from semver import VersionInfo
        >>> VersionInfo.parse('3.4.5-pre.2+build.4')
        VersionInfo(major=3, minor=4, patch=5, \
prerelease='pre.2', build='build.4')

        :return: a :class:`VersionInfo` instance
        :rtype: :class:`VersionInfo`
        """
```

If you change the class name, the test will fail.

scls19fr:
All right! Thanks again @tomschr

PR #124, opened 2019-03-24
Fixes #109 and add iterator to VersionInfo

tomschr:
This PR is based on #109 by Karol Werner and contains:

* Add Karl to the `CONTRIBUTORS` file and added him as primary author.
* Use `version` fixture in test function `test_version_info_should_be_iterable`.
* I've deviated from Karl's test function a bit and avoided the "private" `_astuple()` function and created the tuple manually.
* This PR does not contains a conflict. :wink: 

I think, Karol's idea is useful as it allows to create a list with just `tuple(version)` or `list(version)` without using the (kind of) "private" member function `_astuple()`.

If the maintainers of this project find it useful, I suggest to close #109 as well (as it is related and contains a conflict).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,7 +29,7 @@ Significant contributors
 * Craig Blaszczyk <masterjakul@gmail.com>
 * Jan Pieter Waagmeester <jieter@jieter.nl>
 * Jelo Agnasin <jelo@icannhas.com>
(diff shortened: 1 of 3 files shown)


scls19fr:
Thanks @tomschr and @ppkt for this work

tomschr:
Wow, that was quick! :laughing: Thanks! :+1:

PR #125, opened 2019-03-24
Create setup.cfg for pytest & tox config and coverage

tomschr:

* Use `pytest` command instead of `py.test` (the latter with the dot may be obsolete in the future).
* Move options from `tox.ini` to `setup.cfg`.
* Introduce `pytest-cov` in "tox.ini" for coverage and config file `.coveragerc`
* Add additional options to activate coverage for `semver.py` script and output coverage report

---

Some comments:

* It simplifies the `tox.ini` config. Well, it's probably a matter of taste where do you want to have all your options. However, if you add it to `setup.cfg`, you can also add options for `flake8`. That way, you have everything in **one place**.
* I'm not sure if coverage is useful for such a small script. To have an oversight of which lines are not covered by tests yet *could* be useful. At the moment I get:

```
----------- coverage: platform linux, python 3.6.5-final-0 -----------
Name        Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------
semver.py     186      1     48      3    96%   82, 379, 540-541, 37->42, 81->82, 378->379, 434->438, 539->540
```

Decide yourself if you think this could be useful. If you don't like it, I can change it or just close this PR. :wink:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ envlist =
     pypy
 
 [testenv]
-commands = pytest
+commands = py.test
 deps =
     pytest
     pytest-cov
--- a/tox.ini
+++ b/tox.ini
@@ -5,8 +5,10 @@ envlist =
     pypy
 
 [testenv]
-commands = py.test -q  --doctest-modules
-deps = pytest
+commands = pytest
+deps =
+    pytest
+    pytest-cov
 setenv =
     PIP_DISABLE_PIP_VERSION_CHECK = 1
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+# https://pytest-cov.readthedocs.io/
+
+[run]
+source = semver
+branch = True
+
+[report]
+exclude_lines =
+    pragma: no cover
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -5,8 +5,10 @@ envlist =
     pypy
 
 [testenv]
-commands = py.test -q  --doctest-modules
-deps = pytest
+commands = py.test
```
scls19fr: I wonder if doctests are still running.

Maybe you should create an error in a doctest to ensure that they are still running

tomschr: > I wonder if doctests are still running.

They do. I've tested it locally. The options from the `setup.cfg` file are passed to pytest. As a summary, I get:

```
126 passed in 0.47 seconds
```

tomschr:
Finally, the tests are green! :rocket: Sorry for the commits, I always get hit by Travis. :laughing:

tox.ini
tomschr: According to the [pytest documentation](https://docs.pytest.org/en/latest/customize.html#finding-the-rootdir):

 > "[it] look[s] for pytest.ini, tox.ini and setup.cfg files in the ancestor directory and upwards. If one is matched, it becomes the ini-file and its directory becomes the rootdir."

tomschr:
Thanks! :+1:

scls19fr:
My pleasure!

PR #127, opened 2019-03-24
Fix #126: Provide tox.ini for documentation

### LINKED ISSUE #126: Missing target in tox.ini for building documentation
## Problem

Currently, there is no target inside the `tox.ini` config which can build documentation.

## Expected Solution

It would be helpful if the `tox.ini` config could contain a `doc` section so everybody (including Travis CI) could create documentation.

tomschr:
This PR contains the following change and fixes #126 :

* Add "docs" target to run sphinx and output HTML
* Provide `docs/requirements.txt` file
* Fix typo in `docs/conf.py`

----

Some comments:

* You could run the `docs` target through Travis, same as with the other targets. However, I wasn't sure if this would be useful for you, because you may create the doc with something else.
* I've provided a separate requirements.txt file only for documentation. That way you can add whatever is needed without changing the `tox.ini` file.
* The result of the build can be found at `dist/docs`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -172,6 +172,6 @@ texinfo_documents = [
 ]
 
 
-interpshinx_mapping = {
+intersphinx_mapping = {
     "matplotlib": ('http://matplotlib.org', None),
 }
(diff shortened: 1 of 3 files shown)


tomschr:
Any news on this PR? :wink: Do you need help?

PR #129, opened 2019-04-15
Update doc for format_version to match expected param types

### LINKED ISSUE #128: Docstring for format_version is wrong
https://github.com/k-bx/python-semver/blob/344ea8902a52b198534e15759925884601ac7509/semver.py#L403-L417

major, minor and patch can not be passed as `str`, despite docstring saying so because format specifier expects integers.

zgoda-mobica:
This will fix #128

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -403,9 +403,9 @@ def min_ver(ver1, ver2):
 def format_version(major, minor, patch, prerelease=None, build=None):
     """Format a version according to the Semantic Versioning specification
 
-    :param str major: the required major part of a version
-    :param str minor: the required minor part of a version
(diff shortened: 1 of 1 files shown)


scls19fr:
Thanks @zgoda-mobica for your contribution.

gene1wood:
This may have [broken the build](https://travis-ci.org/k-bx/python-semver/builds/520434813)

scls19fr:
Good catch @gene1wood. I thought [master branch was protected](https://help.github.com/en/articles/about-protected-branches) and merge too quickly!

I did a quick PR #130 to revert code ... but I don't think this problem is really related to this fix.

PR #132, opened 2019-05-29
Fix comparison

Flamefire:
Continuation of stale #104 with fix for Python2

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,14 @@ Python SemVer library
 All notable changes to this code base will be documented in this file,
 in every released version.
 
+Version 2.8.x
+=============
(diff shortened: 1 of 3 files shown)


scls19fr:
@Flamefire could you rebase?

It should
Closes #102
Closes #103

Flamefire:
This should work now. I extended the test case to include all operators in `a <op> b` and `b <op> a`.

Note that `TypeError` must be raised for unsupported comparisons. Otherwise Python assumes "Unimplemented" as in "wanted but not yet done" and does instance comparison as fall back and this is NOT wanted

Flamefire:
Ok, passing now. I cleaned the commits and squashed them into 1.

Can you push out a PyPi release too?

Issue #131, opened 2019-04-16
Travis-CI with Python 3.4 and PyPy is currently broken


scls19fr:
See https://github.com/k-bx/python-semver/pull/129#issuecomment-483786394
and this build https://travis-ci.org/k-bx/python-semver/builds/520902262

Issue #112, opened 2018-10-10
Add Python 3.7 support


scls19fr:
https://github.com/k-bx/python-semver/pull/111#issuecomment-428574713

PR #113, opened 2018-10-10
[WIP] Add Python 3.7 support

scls19fr:

PR #133, opened 2019-05-29
Fix CI

Flamefire:
Old virtualenv uses non-existing pip flags

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
-install: pip install "virtualenv"
+install:
+  - pip install --upgrade pip setuptools
+  - pip install virtualenv
 script: python setup.py test
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,8 @@ matrix:
   - python: "3.6"
     env: TOXENV=py36
 
+  - python: "3.7"
+    env: TOXENV=py36
+
   - python: "pypy"
     env: TOXENV=pypy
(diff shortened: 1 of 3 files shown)


scls19fr:
Thanks @Flamefire for this contribution which seems to fix #131
Maybe you can try to add Python 3.7 support also (and so it could also fix #112 )

Flamefire:
I used the existing PR #113. Will see if it passes

.travis.yml
```diff
@@ -20,5 +20,8 @@ matrix:
   - python: "3.6"
     env: TOXENV=py36
 
+  - python: "3.7"
+    env: TOXENV=py36
```
scls19fr: I wonder why we didn't have

```
  - python: "3.7"
    env: TOXENV=py37
```

Flamefire: C&P I guess. Fixed

scls19fr:
https://github.com/travis-ci/travis-ci/issues/9815 seems to deal with Python 3.7 support with Travis.

Python 3.7 release date is June 27, 2018 ... so I thought that it was supported "out-of-the-box" by Travis but it doesn't seems to be the case (using `xenial` seems to still be necessary)

scls19fr:
LGTM
Thanks @Flamefire

Flamefire:
Using xenial is the official way for 3.7: https://docs.travis-ci.com/user/languages/python/#python-37-and-higher

PR #138, opened 2019-08-16
Add `__getitem__` method to allow easy access to version parts

tlaferriere:
This is a natural extension of the work done in #124 (and parents) and makes it easy to access version parts using their index.  

##### Use case:
Using an `Enum` to access the version part:
```python
VersionPartEnum = Enum('VersionPartEnum', "MAJOR MINOR PATCH PRERELEASE BUILD")
version_info = VersionInfo.parse("1.2.3-rc.1+build.13")

print(version_info[VersionPartEnum.MINOR - 1])  # 2
```


scls19fr:
Thanks @tlaferriere for this PR
but a feature should never be implemented without unit tests.

tlaferriere:
I'll get on it.

tlaferriere:
@scls19fr Should I add test vectors for edge cases or are these ones satisfactory?

scls19fr:
I would be pleased to have the general consensus about this feature before merging
Pinging @k-bx @kxepal @bittner @tomschr @ppkt

semver.py
```diff
@@ -164,6 +164,9 @@ def __iter__(self):
         for v in self._astuple():
             yield v
 
+    def __getitem__(self, index):
+        return self._astuple()[index]
```
ppkt: Are we going to support negative index? e.g. `version_info[-3]`? In my opinion - no - so we should raise an `IndexError` exception if index is less than zero.

test_semver.py
```diff
@@ -497,3 +497,15 @@ def test_immutable_unknown_attribute(version):
...
+@pytest.mark.parametrize("version, index, expected", [
+    ("1.2.3-rc.0+build.0", 0, 1),
+    ("1.2.3-rc.0+build.0", 1, 2),
+    ("1.2.3-rc.0+build.0", 2, 3),
+    ("1.2.3-rc.0+build.0", 3, "rc.0"),
+    ("1.2.3-rc.0+build.0", 4, "build.0"),
+])
+def test_version_info_should_be_accessed_with_index(version, index, expected):
+    version_info = VersionInfo.parse(version)
+    assert version_info[index] == expected
```
ppkt: I would add more tests for corner case scenarios - for instance - what if are trying to get index of prerelease part but `VersionInfo` does not have such field? e.g.
`assert VersionInfo.parse("1.2.3")[4] == None`

Also, please add tests for negative and too big index value:
`assert VersionInfo.parse("1.2.3-rc.0+build.0")[-1] == "build.0"` and
`with pytest.raises(IndexError): VersionInfo.parse("1.2.3-rc.0+build.0")[5]`

bittner:
The implementation is certainly fine.

I'm not super-convinced by the above use case, though. While it's a fact that Enums start at 1 [for a reason](https://docs.python.org/3/library/enum.html#functional-api) I'd prefer to see a more intuitive syntax (i.e. something that doesn't need to do index calculations, namely `- 1`). Then, if this is a real use case, why do we require the user to come up with the needed Enum?

The [doctests of parse_version_info](https://github.com/k-bx/python-semver/blob/master/semver.py#L233-L244) may also include access with constants, and we may provide the constants directly, e.g.

```python
>>> import semver
>>> version_info = semver.parse_version_info("3.4.5-pre.2+build.4")
>>> version_info.major == version_info[semver.MAJOR]
True
>>> version_info.minor == version_info[semver.MINOR]
True
...
```
`MAJOR` & friends would have to be (read-only) class member for this to work.

semver.py
tlaferriere: It is not very hard to support negatives indices, as Python already provides this functionality with tuples. On the one hand, we can stay coherent with the reality of a version string in the way that we don't think of it in terms of a sequence, making negative indices unnecessary and incoherent.   

On the other hand, most python developers use negative indices and slices because this syntactic sugar is very convenient and intuitive in most cases. One could argue that using negative indices would be part of muscle memory in an experienced python developer's hands, thus allowing the use of negative indices more intuitive and natural.  

But the reality is that negative indices should be used only for sequences of variable length. Our version sequence length, in the state the project is now (correct me if I'm wrong), can only vary by 2 elements (prerelease and build number). *Can we consider this variable enough to necessitate negative indices?*

From what I gleaned by skimming over the issues, there are proposals to support incomplete version strings (like `major` or `major.minor`). If that were to be realised, negative indices could indeed become more useful. Unless other features are implemented that would make this redundant (a way to detect the *partialness* of a version that doesn't involve negative indices.)

Right now the path of least resistance is to allow the use of negative indices, since we are using a tuple internally to represent the version. In the future, we could remove support for this by adding the necessary checks, but that would break functionality that has been implemented, which would require a major version release according to semver.

I believe I laid out most of the arguments for and against negative indexing. In my opinion, we should go the path of least resistance, at the risk of having to break functionality in the future. I consider this risk relatively low, because negative indexing could become useful internally to this project.

I would like to know what the consensus is on this feature (or bug ;) ).

test_semver.py
tlaferriere: I'll add test vectors for these cases.

tomschr:
@bittner I like your idea! :+1: :smiley:

semver.py
tomschr: Here are my 2 cents:

> But the reality is that negative indices should be used only for sequences of variable length. Our version sequence length, in the state the project is now (correct me if I'm wrong), can only vary by 2 elements (prerelease and build number). _Can we consider this variable enough to necessitate negative indices?_

IMHO, I think, that is the main argument against using negative indices. 

> Right now the path of least resistance is to allow the use of negative indices, since we are using a tuple internally to represent the version.

True, but from an OOP perspective, this is an implementation detail; the class should be perceived as a "black box". The user doesn't have to know how the class works internally.

---
What about the following strategy: If users doesn't ask for this specific feature, why implementing it? It there is no need, we can spare the work and focus on something different. Of course, we can come back, if people requesting this feature.

Also, **don't forget to document it** (regardless if this is implemented with or without negative indices).

test_semver.py
tomschr: > I would add more tests for corner case scenarios - for instance - what if are trying to get index of prerelease part but `VersionInfo` does not have such field? e.g.
> `assert VersionInfo.parse("1.2.3")[4] == None`
> 
> Also, please add tests for negative and too big index value:
> `assert VersionInfo.parse("1.2.3-rc.0+build.0")[-1] == "build.0"` and
> `with pytest.raises(IndexError): VersionInfo.parse("1.2.3-rc.0+build.0")[5]`

That's a good idea! May I suggest to add the tests for a negative index value as a separate test function?
I think, that makes it a bit easier as you don't have to change `test_version_info_should_be_accessed_with_index` (as test function should be short).

scls19fr:
This PR have a conflicting file (sorry I should have managed this one before but was and still is quite busy).
Please resolve conflicts and thumb up if you are ok to merge this.
Thumb down if you don't want this PR to be merged.

scls19fr:
Please fix this conflict to have a chance to have this feature include in latest release!

tlaferriere:
Excuse the delay, I am resolving the conflicts and finishing up tests.

tlaferriere:
Also I've decided not to add the `MAJOR`, `MINOR`, etc. members as it feels redundant with the key indices like `"major"`, `"minor"`, etc. After all, the goal was to access it like an iterable, and not a dictionary.

scls19fr:
Thanks @tlaferriere for this first contribution here.

Could you update doc to show this new feature?
I wonder if we shouldn't have example in docstring also.

Could you also update (CHANGELOG)[https://github.com/k-bx/python-semver/blob/master/CHANGELOG.rst]

scls19fr:
You can also add your name to https://github.com/k-bx/python-semver/blob/master/CONTRIBUTORS

semver.py
```diff
@@ -168,6 +168,12 @@ def __iter__(self):
         for v in self._astuple():
             yield v
 
+    def __getitem__(self, index):
+        """Implement getitem. This automatically strips empty parts of the
+           version from the iterable from which the index is taken."""
```
tomschr: Could you change:

```suggestion
           version from the iterable from which the index is taken.

           :param int index: a positive or negative integer
           :raises: IndexError, if index is beyond the range
           """
```

ppkt: @tomschr according to current implementation, index can be negative.

@tlaferriere I think there is a subtle bug in your implementation:
```
    def test_example():
        v = VersionInfo.parse('1.2.3+build0')
>       assert v[:] == (1, 2, 3, None, 'build0')
E       AssertionError: assert (1, 2, 3, 'build0') == (1, 2, 3, None, 'build0')
E         At index 3 diff: 'build0' != None
E         Right contains more items, first extra item: 'build0'
E         Use -v to get the full diff
```
i.e. if "prerelease" part is missing and "metadata" is present, after using __getitem__ , "metadata" appears at index for "prerelease"

tomschr: @ppkt 
> according to current implementation, index can be negative.

You are right, I've corrected the GH suggestion. Thanks! :+1:

scls19fr:
Conflicting files need to be solved here also. CI is failing and should be fixed also.

semver.py
tlaferriere: After much deliberation with myself and an increasingly complex function, I have decided to rewrite it (multiple times). I believe I have made this implementation as simple as possible. Turns out the subtle bug mentioned by @ppkt was a symptom of a few bad design decisions on my part. Now it doesn't support negative indices.

semver.py
```diff
@@ -174,6 +174,39 @@ def __iter__(self):
...
 
+    def __getitem__(self, index):
+        """Implement getitem. If the part requested is undefined, or a part of the
+           range requested is undefined, it will throw an index error.
+           Negative indices are not supported
+
+           :param Union[int, slice] index: a positive integer indicating the
+                  offset or a slice
+           :raises: IndexError, if index is beyond the range or a part is None
+           """
```
bittner: This docstring has an unusual indentation. According to [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) indentation may look like:
```python
def foo():
    """First line of comment

    Second line of comment. The longer description.
    """
```
(what I like less but seems to be favored by Guido & friends), or:
```python
def foo():
    """
    First line of comment

    Second line of comment. The longer description.
    """
```

Same below, in the method following next.

semver.py
```diff
@@ -174,6 +174,39 @@ def __iter__(self):
...
+        undefined_error = IndexError("Version part undefined")
+        negative_error = IndexError("Version index cannot be negative")
+        if isinstance(index, slice):
+            if (False if index.start is None else index.start < 0) or \
+                    (False if index.stop is None else index.stop < 0):
+                raise negative_error
+            else:
+                slice_is_full = True
+                part = self._astuple()[index]
+                for i in part:
```
bittner: What is `i` here? Pylint would be complaining (if you ran it). Can you use a self-explanatory name?

semver.py
```diff
@@ -174,6 +174,39 @@ def __iter__(self):
...
+                        slice_is_full = False
+                    elif not slice_is_full:
+                        raise undefined_error
+                part = tuple(filter(None, part))
+        else:
+            if index < 0:
+                raise negative_error
+            else:
+                part = self._astuple()[index]
+                if part is None:
```
bittner:
```python
if not part:
```
... is probably more pythonic. Or is there a reason for an explicit `None`?

bittner: This is in several places in the code, by the way, and could be made pythonic throughout the code.

tomschr:
@tlaferriere Found another issue.

You could simplify the code a bit and omit the `else` branches:

```python
def __getitem__(self, index):
        """
           Implement getitem. If the part requested is undefined, or a part of the
           range requested is undefined, it will throw an index error.
           Negative indices are not supported

           :param Union[int, slice] index: a positive integer indicating the
                  offset or a slice
           :raises: IndexError, if index is beyond the range or a part is None
           """
        undefined_error = IndexError("Version part undefined")
        negative_error = IndexError("Version index cannot be negative")
        if isinstance(index, slice):
            if (False if index.start is None else index.start < 0) or \
                    (False if index.stop is None else index.stop < 0):
                raise negative_error
            slice_is_full = True
            part = self._astuple()[index]
            for i in part:
                if i is None:
                    slice_is_full = False
                elif not slice_is_full:
                    raise undefined_error
            part = tuple(filter(None, part))
        else:
            if index < 0:
                raise negative_error
            part = self._astuple()[index]
            if part is None:
                raise undefined_error
        return part
```

semver.py
tomschr: One idea could be to replace `i` with `item`.

tomschr:
I could find a more optimized version which (IMHO) is even more readable. It reduces the line count from 21 to 14. It also avoids the for-loop:

```python
def __getitem__(self, index):
        """Implement getitem. If the part requested is undefined, or a part of the
           range requested is undefined, it will throw an index error.
           Negative indices are not supported

           :param Union[int, slice] index: a positive integer indicating the
                  offset or a slice
           :raises: IndexError, if index is beyond the range or a part is None
           :return: the requested part of the version at position index
        """
        undefined_error = "Version part undefined"
        negative_error = "Version index cannot be negative"

        if isinstance(index, int):
            index = slice(index, index+1)

        if isinstance(index, slice) and \
                ((False if index.start is None else index.start < 0) or \
                 (False if index.stop is None else index.stop < 0)):
            raise IndexError(negative_error)

        # Could raise IndexError:
        part = tuple(filter(None, self._astuple()[index]))

        if len(part) == 1:
            part = part[0]
        if not part:
            raise IndexError(undefined_error)
        return part
```

tomschr:
@tlaferriere Thomas, what's your opinion on https://github.com/python-semver/python-semver/pull/138#issuecomment-557935244?

Could we finish this PR to publish it for the 2.9.1 release or do you need more time? :wink:  If you don't have time yet due to the holiday season, I could finish it for you if you like. Let me know what you prefer. :wink:

semver.py
tlaferriere: I guess I could use `subpart` instead.

tomschr: See also the more recent comments below. :wink:

tlaferriere:
@tomschr with my finals ending soon, I figured I would have more time to finish it up and get it ready for the release. I like the snippet, I haven't had time to inspect it, but it sure seems cleaner. I'll get some work in the next few days and hopefully it will go well enough for it to be ready.

tomschr:
Take your time. If you need any help, let us know. :blush:

tomschr:
@tlaferriere Thomas, just a quick question: if you are too busy at the moment, should I finish the issue for you or would you prefer to do it yourself? Of course I will add you as co-author. :wink:

tomschr:
I've created this PR anew in #243, so closing this.

PR #140, opened 2019-09-29
Fix #135, convert prerelease and build to string

### LINKED ISSUE #135: Convert arguments in VersionInfo ctor
It just happened to me to use `VersionInfo(3, 2, 1, 1)` which later failed due to the prerelease not being a string which the comparison assumed.

I propose to **either**
1. Convert arguments in ctor of `VersionInfo` to their correct types (int/string)
2. Assert the correct type

tomschr:
This PR fixes #135 

Creating `VersionInfo(3, 2, 1, 1)` succeeds, but fails later when comparing it with another `VersionInfo` instance (for example, v1 < v2):

```
    def split_key(key):
>       return [convert(c) for c in key.split('.')]
E       AttributeError: 'int' object has no attribute 'split'
```

### DIFF
--- a/test_semver.py
+++ b/test_semver.py
@@ -499,13 +499,9 @@ def test_version_info_should_be_iterable(version):
                               version.prerelease, version.build)
 
 
-@pytest.mark.parametrize("v1,v2", [
-  # no. 1
-  ({'major': 1, 'minor': 9, 'patch': 1, 'prerelease': 1, 'build': 1},
-   {'major': 1, 'minor': 9, 'patch': 1, 'prerelease': 2, 'build': 1}),
-  # no. 2
-  ({'major': 3, 'minor': 2, 'patch': 1, 'prerelease': 1},
-   {'major': 3, 'minor': 2, 'patch': 1, 'prerelease': 3}),
-])
-def test_should_compare_prerelease_and_build_with_numbers(v1, v2):
-    assert VersionInfo(**v1) < VersionInfo(**v2)
+def test_should_compare_prerelease_and_build_with_numbers():
+    assert VersionInfo(major=1, minor=9, patch=1, prerelease=1, build=1) < \
+           VersionInfo(major=1, minor=9, patch=1, prerelease=2, build=1)
+    assert VersionInfo(1, 9, 1, 1, 1) < VersionInfo(1, 9, 1, 2, 1)
+    assert VersionInfo('2') < VersionInfo(10)
+    assert VersionInfo('2') < VersionInfo('10')
--- a/test_semver.py
+++ b/test_semver.py
@@ -514,10 +514,11 @@ def test_should_be_able_to_use_strings_as_major_minor_patch():
     assert isinstance(v.patch, int)
     assert v.prerelease is None
     assert v.build is None
+    assert v == VersionInfo(1, 2, 3)
+    assert VersionInfo(1, 2, 3, 4, 5) == VersionInfo(1, 2, 3, '4', '5')
 
 
 def test_should_be_able_to_use_integers_as_prerelease_build():
     v = VersionInfo('1', '2', '3', 4, 5)
     assert isinstance(v.prerelease, str)
     assert isinstance(v.build, str)
-    assert VersionInfo('1', '2', '3') == VersionInfo(1, 2, 3)
--- a/test_semver.py
+++ b/test_semver.py
@@ -514,11 +514,20 @@ def test_should_be_able_to_use_strings_as_major_minor_patch():
     assert isinstance(v.patch, int)
     assert v.prerelease is None
     assert v.build is None
-    assert v == VersionInfo(1, 2, 3)
-    assert VersionInfo(1, 2, 3, 4, 5) == VersionInfo(1, 2, 3, '4', '5')
+    assert VersionInfo('1', '2', '3') == VersionInfo(1, 2, 3)
+
+
+def test_using_non_numeric_string_as_major_minor_patch_throws():
+    with pytest.raises(ValueError):
+        v = VersionInfo('a')
+    with pytest.raises(ValueError):
+        v = VersionInfo(1, 'a')
+    with pytest.raises(ValueError):
+        v = VersionInfo(1, 2, 'a')
 
 
 def test_should_be_able_to_use_integers_as_prerelease_build():
-    v = VersionInfo('1', '2', '3', 4, 5)
+    v = VersionInfo(1, 2, 3, 4, 5)
     assert isinstance(v.prerelease, str)
     assert isinstance(v.build, str)
+    assert VersionInfo(1, 2, 3, 4, 5) == VersionInfo(1, 2, 3, '4', '5')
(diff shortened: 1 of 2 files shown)


test_semver.py
```diff
@@ -497,3 +497,9 @@ def test_immutable_unknown_attribute(version):
 def test_version_info_should_be_iterable(version):
     assert tuple(version) == (version.major, version.minor, version.patch,
                               version.prerelease, version.build)
+
+
+def test_should_compare_prerelease_and_build_with_numbers():
+    v1 = VersionInfo(major=1, minor=9, patch=1, prerelease=1, build=1)
+    v2 = VersionInfo(major=1, minor=9, patch=1, prerelease=2, build=1)
```
Flamefire: Minor: I suggest to add a test with ` VersionInfo(3, 2, 1, 1)` style ctor as used in the commit description.

tomschr: @Flamefire Thanks for the suggestion! :+1:  I've added this in the next commits (see below).

tomschr: > Doesn't `**v1` expand the dictionary to kwargs? If so then there is no `VersionInfo(3, 2, 1, 1)` usage. 

Well, yes, you could replace the dictionary with a tuple. Would be the same. :wink: 


> PS: I would not use the `parametrize` feature here. It isn't really shorter and obscures the meaning more than it helps (see my confusion above). Why not put a couple asserts in the tests instead? 

I think you are right, the test could be simplified.

> This would also double as documentation:
> 
> ```
> assert VersionInfo(major=1, minor=9, patch=1, prerelease=1, build=1) <  VersionInfo(major=1, minor=9, patch=1, prerelease=2, build=1)
> assert VersionInfo(1, 9, 1, 1, 1) <  VersionInfo(1, 9, 1, 2, 1)
> assert VersionInfo(10) <  VersionInfo('2')
> assert VersionInfo('10') <  VersionInfo('2')
> ```

I think, the last two conditions are wrong. At least, the test function gives me an AssertionError. I think what you mean is this:

```python
assert VersionInfo('2') <  VersionInfo(10)
assert VersionInfo('2') <  VersionInfo('10')
```

Correct? Can you confirm?

Flamefire: Yes correct of course, C&P ;-) I also mistakenly added a double space, so make sure to remove it on the patch

tomschr: Haha, ok. No problem. I will add it and you as a co-author. Thanks! :+1:

tomschr: @Flamefire  Hmn, after looking at the function name of the test case, it seems it doesn't really show its intention. Unfortunately, I didn't have a better idea.

Maybe do you have a better name? :wink:  Or any other things I should change?

Flamefire: Maybe: Split the 2nd part into a new test like `test_should_be_able_to_use_strings_as_major_minor_patch` and add additional asserts for the minor and patch. Maybe it is even enough to do:

```
v = VersionInfo('1', '2', '3')
assert v.major == 1
assert isinstance(v.major, int)
...
assert v.prerelease is None
assert v.build is None
```

And a similar test for the int->string in prerelease&build

tomschr: Good idea! :+1: I have one question regarding this code:

```python
assert v.major == 1
assert isinstance(v.major, int)
```

Could we avoid the first assert? The reason why I'm asking is: this is covered already in other tests. IMHO, we don't need to repeat ourselves. In this test we are interested in the correct *type* only, not the correct *content*, right?

If you agree, it would simplify the test a bit and focus on the type check with `isinstance`. With that, I end up with this test:

```python
def test_should_be_able_to_use_strings_as_major_minor_patch():
    v = VersionInfo('1', '2', '3')
    assert isinstance(v.major, int)
    assert isinstance(v.minor, int)
    assert isinstance(v.patch, int)
    assert v.prerelease is None
    assert v.build is None
```

Ok with that? :-)

Flamefire: I would add an assertion `VersionInfo('1', '2', '3') == VersionInfo(1, 2, 3)` so a typo like ` self._minor = int(major)` is caught. While I agree that this tests the *type* it must also assure that the conversion is right. Maybe someone implements it as `int(major) if isintance(major, string) else major` later. This is not covered by any other test. The above assertion is also very good for documentation as it is as descriptive as it gets.

tomschr: Yes, you're right. I've added your additional line and my above lines in commit a611a73. Better? :wink:

test_semver.py
```diff
@@ -505,3 +505,19 @@ def test_should_compare_prerelease_and_build_with_numbers():
...
+    assert isinstance(v.patch, int)
+    assert v.prerelease is None
+    assert v.build is None
+
+
+def test_should_be_able_to_use_integers_as_prerelease_build():
+    v = VersionInfo('1', '2', '3', 4, 5)
+    assert isinstance(v.prerelease, str)
+    assert isinstance(v.build, str)
+    assert VersionInfo('1', '2', '3') == VersionInfo(1, 2, 3)
```
Flamefire: Move this to `test_should_be_able_to_use_strings_as_major_minor_patch` and add here: `assert VersionInfo(1, 2, 3, 4, 5) == VersionInfo(1, 2, 3, '4', '5')`. Also use integers in the ctor at line 520 as this is about the prerelease and build members to be "unconventional" types not the first 3.

test_semver.py
Flamefire: Added a comment  on the line with suggestion :)

test_semver.py
tomschr: > Move this to `test_should_be_able_to_use_strings_as_major_minor_patch` and add here: `assert VersionInfo(1, 2, 3, 4, 5) == VersionInfo(1, 2, 3, '4', '5')`. 

Ok, I did this in commit 9361751. However, the term "here" is ambiguous. :wink: Better check the commit.

> Also use integers in the ctor at line 520 as this is about the prerelease and build members to be "unconventional" types not the first 3.

Not sure if I got it. The `VersionInfo.__init__` already have an `int()` and `str()` call. Maybe show it with an example what you mean?

Flamefire: My suggestion would be (fix the '???'):

```
def test_should_be_able_to_use_strings_as_major_minor_patch():
    v = VersionInfo('1', '2', '3')
    assert isinstance(v.major, int)
    assert isinstance(v.minor, int)
    assert isinstance(v.patch, int)
    assert v.prerelease is None
    assert v.build is None
    assert VersionInfo('1', '2', '3') == VersionInfo(1, 2, 3)


def test_using_non_numeric_string_as_major_minor_patch_throws():
    with pytest.raises('???'):
        v = VersionInfo('a')
    with pytest.raises('???'):
        v = VersionInfo(1, 'a')
    with pytest.raises('???'):
        v = VersionInfo(1, 2, 'a')


def test_should_be_able_to_use_integers_as_prerelease_build():
    v = VersionInfo(1, 2, 3, 4, 5)
    assert isinstance(v.prerelease, str)
    assert isinstance(v.build, str)
    assert VersionInfo(1, 2, 3, 4, 5) == VersionInfo(1, 2, 3, '4', '5')
```

The last line in the changes tests is very explicit on purpose to serve as easily readable documentation. Important change is the focus on what is tested in each test and addition of a "corner case" test

tomschr: Thanks for the code, very appreciated! I've covered the `???`, it's a `ValueError` exception. I've copied and adapted your code in commit 9361751. Many thanks! :+1:

scls19fr:
Any news about this PR? Is there a general consensus?

tomschr:
From my side it's ok, :wink:  but would @Flamefire approve? :wink:

scls19fr:
Ok. So let's merge it.
Thanks guys for your work!

Issue #139, opened 2019-09-29
Bump version inside VersionInfo directly


tomschr:
## Situation

Hope I don't miss anything, but reading the source code didn't give a solution. It seems to me, dealing with version numbers and raising parts of it can actually quite tricky.

For example, let's assume you've created a version like this:

```python
ver = semver.parse_version_info("3.4.5")
```

How can you raise the major part? How the minor part? The `ver` variable is of type `VersionInfo`. You can't directly use `semver.bump_major` or `semver.bump_minor` as they accept strings only. Nor does the class provide any methods. However,  you *could* do something like that:

```python
semver.bump_major(str(ver))
```

This works, but your `VersionInfo` type is converted into a string. Not good if you want to keep the type. If you want it back, you need to write:

```python
semver.parse_version_info(semver.bump_major(str(ver)))
```

Not very pythonic. :wink: 

## Possible Solution(s)

From what I can see, we have a good class, but we can't raise parts individually. I think, this is gap. I could think of these solutions:

1. Allow `semver.bump_major` and the other bump functions allow a `VersionInfo` type.
2. Extend the `VersionInfo` class with additional methods like `inc_major()`, `inc_minor()` etc.
3. Something else?

At the moment, I think number 2 can be a possible solution. Theoretically, you could name them also `bump_major`, but that could confuse people.

The above could be rewritten like that:

```python
ver.inc_major()
```

If we go with this, there are two possible ways how this function could work:

1. Return the raised version of `ver` but leave `ver` itself untouched.
2. Return nothing (=None) and raise the part directly in `ver`.

Solution 1 could be interesting as it would allow to write something like this:

```python
new_ver = ver.inc_major().inc_minor()
```

What do you think?

ppkt:
@tomschr you're right, currently we don't have such feature :)

Personally, I like OO approach here and I think adding these functions would be really nice. Regarding your proposal - I like options 2 and 1 respectively (but I'd like to stick with `bump_minor` to match name with functions working on strings). As alternative, we can add new function as class method to avoid confusion regarding whether function returns something or modify object itself:
`new_ver = VersionInfo.bump_major(ver)`

tomschr:
> @tomschr you're right, currently we don't have such feature :)

Ahh good, that means, I didn't overlook anything. :wink: 


> Personally, I like OO approach here and I think adding these functions would be really nice. 

Mee too.

[...]

Issue #40, opened 2016-11-28
semver-object


Thor77:
In my opinion your approach to parsing is quite "unpythonic", I would prefer smth like this:
```py
info = semver('1.0.0')
```
that would allow easy comparison with
```py
>> semver('1.0.0') > semver('2.0.0')
>> False
```
(yes I know there's `semver.compare` for that)
Python is a object-orientated language after all ;)

k-bx:
@Thor77 you can already do that:

```
In [3]: info1 = semver.parse_version_info('1.0.0')

In [4]: info2 = semver.parse_version_info('2.0.0')

In [5]: info1 > info2
Out[5]: False
```

Thor77:
I know there are already ways to accomplish this in your library, but they're verbose and not object-orientated.
For example
```py
ver = semver.parse_version_info('1.0.0')
ver.bump_{major,minor,patch}()
```
does not work.

k-bx:
@Thor77 ah, I see what you mean. So, if you want to make a nice PR implementing a Semver class with all the object-oriented functionality, I wouldn't mind accepting it :)

scls19fr:
@Thor77 I think after PR #88 and #141 we can close this (quite old) issue!
Please feel to reopen if you think that's necessary.

PR #141, opened 2019-09-30
Implement #139: Bump version inside VersionInfo

tomschr:
This PR implements #139 

* Add bump_major, bump_minor, bump_patch, bump_prerelease and bump_build as methods in class VersionInfo. With this methods, it is now possible to write something like this:

    ```python
    ver = semver.parse_version_info("3.4.5")
    new_ver = ver.bump_major().bump_minor()
    ```

* Add test cases

---

Some comments about this implementation:

* I (re)used the module functions `semver.bump_major()` etc. to avoid implementing the same code (double code is bad code).
* It may be not as efficient as it could be, because it converts the VersionInfo into a string, bumps it, and convert is back into VersionInfo. Not sure if there is a better way to do it.
* I've created a "Draft pull request" so we can discuss and improve the code. ;-)

@ppkt Karol, what do you think? :-)

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -164,6 +164,74 @@ class VersionInfo(object):
         for v in self._astuple():
             yield v
 
+    def bump_major(self):
+        """Raise the major part of the version
+
+        :return: the raised version string
+        :rtype: VersionInfo
+
+        >>> import semver
+        >>> ver = semver.parse_version_info("3.4.5")
+        >>> ver.bump_major()
+        VersionInfo(major=4, minor=0, patch=0, prerelease=None, build=None)
+        """
+        return parse_version_info(bump_major(str(self)))
+
+    def bump_minor(self):
+        """Raise the minor part of the version
+
+        :return: the raised version string
+        :rtype: VersionInfo
+
+        >>> import semver
+        >>> ver = semver.parse_version_info("3.4.5")
+        >>> ver.bump_minor()
+        VersionInfo(major=3, minor=5, patch=0, prerelease=None, build=None)
+        """
+        return parse_version_info(bump_minor(str(self)))
+
+    def bump_patch(self):
+        """Raise the patch part of the version
+
+        :return: the raised version string
+        :rtype: VersionInfo
+
+        >>> import semver
+        >>> ver = semver.parse_version_info("3.4.5")
+        >>> ver.bump_patch()
+        VersionInfo(major=3, minor=4, patch=6, prerelease=None, build=None)
+        """
+        return parse_version_info(bump_patch(str(self)))
+
+    def bump_prerelease(self, token='rc'):
+        """Raise the prerelease part of the version
+
+        :param token: defaults to 'rc'
+        :return: the raised version string
+        :rtype: str
+
+        >>> import semver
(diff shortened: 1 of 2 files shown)


semver.py
```diff
@@ -164,6 +164,74 @@ def __iter__(self):
         for v in self._astuple():
             yield v
 
+    def bump_major(self):
+        """Raise the major part of the version
```
ppkt: I would emphasize in docstring that new object is returned instead of modifying `self`

semver.py
```diff
@@ -164,6 +164,74 @@ def __iter__(self):
...
+
+        :return: the raised version string
```
ppkt: maybe "new object with raised major part" would be better? Just to indicate that after calling this function new VersionInfo object is returned, not text string. I know that return type is already described below but I don't know how to do it better :)

semver.py
tomschr: Good catch! :+1: What about this phrase:

```
Raise the major part of the version, return a new object
but leave self untouched
```

semver.py
tomschr: I absolutely agree, it was badly worded. :wink: This part is crucial and needs to be documented.

Will change that.

tomschr:
@ppkt I've implemented your suggestion in the above commits. Hope you are fine with that. :wink: Let me know if you prefer to squash the commits.

By the way: what amount of lines do I need to change to qualify for the contributors file? :wink:

tomschr:
Another question about consistency.

I've found out that some docstring are written differently. Some docstrings have this style:

```python
def parse(version):
    """Parse version to major, minor, patch, pre-release, build parts.

    :return: dictionary with the keys 'build', 'major', 'minor', 'patch',
             and 'prerelease'. The prerelease or build keys can be None
             if not provided
    :rtype: dict

    >>> import semver
    >>> ver = semver.parse('3.4.5-pre.2+build.4')
    >>> # [...]
    """
```

while others have this style:

```python
def bump_major(self):
        """Raise the major part of the version, return a new object
           but leave self untouched

        :return: new object with the raised major part
        :rtype: VersionInfo

        >>> import semver
        >>> ver = semver.parse_version_info("3.4.5")
        >>> ver.bump_major()
        VersionInfo(major=4, minor=0, patch=0, prerelease=None, build=None)
        """
```

Probably that was me to blame. :cry:  To summarize: either we write doctests first and then all parameters, or vice versa. Is there any preferred style? 

IMHO, at least we should be consistent. I would like to fix that, but that's probably something for a different PR, right?

semver.py
ppkt: Yes, it's very clear now, thanks :)

scls19fr:
Thanks @tomschr for this PR and @ppkt for this review.

ppkt:
@tomschr I think you can update docstrings in separate PR. Regarding "contributors" file - I don't think there is any limit - feel free to add yourself. And thank you for your PRs :)

tomschr:
@ppkt Ok, will attack consistency of docstrings in another PR.  :wink:  Regarding contributors file, I will add myself in one of my open PRs. :smile:  Thanks!

scls19fr:
This PR helps this quite old issue: #40

PR #143, opened 2019-09-30
Doc: improve Usage section

### LINKED ISSUE #142: Doc: Improve Usage Section
## Situation

Currently, the documentation in the "Usage" section is limited. It covers the important parts very briefly, but also lacks other features that are added recently, For example, iterating over a `VersionInfo` or passing strings to `VersionInfo` and others.

IMHO, the documentation is an integral and crucial part of helping our users to use the module successfully and efficiently.

## Proposed Solution

The "Usage" section should cover, as an example, but not limited to:

* create instances of `VersionInfo`
* access different parts of a version
* convert different types into each other (for example, a dict into `VersionInfo` and vice versa)
* increase/bump parts of a version
* compare two versions

tomschr:
This PR fixes #142. I know, the PR is quite extensive, but it basically boils down to the `usage.rst` file:

* Use descriptive titles (verb+ing) to better find the topic
* Structure the content with (sub)sections
* Try to structure the different sections with list if there are different options to make it more readable
* Document other, possible unknown features

I also changed the `index.rst` file as I think it can be simplified a bit as well (see commit 
4d15b25 for the details). Hopefully the comments in the commits make it clear what I did.

I recommend to build the doc with `tox -e docs` and to look into the HTML pages.

What do you think? :wink:


tomschr:
It seems, Travis CI complains about a flake issue that appears in Python 3.4 only. The respective error message (line breaks introduced by me) looks like this:

```
flake8 installed: DEPRECATION: Python 3.4 support has been deprecated. 
pip 19.1 will be the last one supporting it. 
Please upgrade your Python as Python 3.4 won't be maintained after
March 2019 (cf PEP 429)
```

scls19fr:
Should we drop Python 3.4 support in this library for this reason?

tomschr:
> Should we drop Python 3.4 support in this library for this reason?

That depends on if you still want to support version 3.4. :wink: If a significant amount of your user base still use it, then it would probably a bad idea. But maybe you should "advertise" it first, for example, in an issue and in the changelog.

Funny side note: I've try to check when version 3.4 was released. I even can't select it in https://docs.python.org from the drop-down menu. :wink: 

If you still want to support version 3.4, you could have replace flake8 with another linter/checker. For example, [pystylecode](https://pycodestyle.readthedocs.io/), [pylint](https://pylint.readthedocs.io/) and probably many others are available.

ppkt:
@tomschr I think problem is somewhere else, from logs it's an import statement: https://travis-ci.org/k-bx/python-semver/jobs/591505916#L301 . In my opinion you can just silence it:
```
from semver import __version__  # noqa: E402
```

Regarding Python 3.4 - I think we should support it as long as it's possible. While I love using new features, probably there is a lot of users who are still using older Python versions (especially 2.7 which will reach EOL status in 3 months: https://pythonclock.org/)

tomschr:
@ppkt Ok Karol, I've added the respective line on the import to silence flake. Let's hope it works. :wink: 

> Regarding Python 3.4 - I think we should support it as long as it's possible. While I love using new features, probably there is a lot of users who are still using older Python versions (especially 2.7 which will reach EOL status in 3 months: https://pythonclock.org/)

I fully agree. :+1:

scls19fr:
Thanks @tomschr for this other nice PR

PR #146, opened 2019-09-30
Fix #145: add posargs in tox.ini (single test cases)

### LINKED ISSUE #145: add posargs in tox.ini (single test cases)
## Situation

Usually you run `tox` like this:

```shell
tox -e py36
```

That works, all fine. However, it runs the *complete* test suite. If you work on a single test function, it is currently not possible to run this alone. You always have to run the complete test suite.

## Proposed Solution

I would propose the following small patch to introduce the `posargs` argument:

```diff
diff --git c/tox.ini i/tox.ini
index e172f8a..f7b1b2e 100644
--- c/tox.ini
+++ i/tox.ini
@@ -5,7 +5,7 @@ envlist =
     pypy
 
 [testenv]
-commands = py.test
+commands = py.test {posargs:}
```

This makes it possible to run `tox` only for the test function `test_should_bump_major`:

```
tox -e py36 test_semver.py::test_should_bump_major
```

With this single patch, it is even possible to use it for the `setup.py` script:

```
python3 setup.py test -a "-e py36 test_semver.py::test_should_bump_major"
```

tomschr:
This PR fixes #145 

Without the posargs argument tox has to run the *complete* test suite. With this patch, you can
pass a test file and a test function. That allows tox to only run this specific test function. For example:

```
$ tox -e py36 test_semver.py::test_should_bump_major
```

Keep in mind, it doesn't change the default behavior. It's just an additional feature if someone wants to run a single test function alone.

This may help to speed up the development a little bit. :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
+# config file for automatic testing at travis-ci.org
 language: python
+cache: pip
 install:
   - pip install --upgrade pip setuptools
   - pip install virtualenv tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 install:
   - pip install --upgrade pip setuptools
   - pip install virtualenv
-script: python setup.py test
+script: tox -v
 matrix:
   include:
   - python: "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 install:
   - pip install --upgrade pip setuptools
-  - pip install virtualenv
+  - pip install virtualenv tox
 script: tox -v
 matrix:
   include:
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ envlist =
     pypy
 
 [testenv]
-commands = py.test {posargs:}
+commands = pytest {posargs:}
 deps =
     pytest
     pytest-cov
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ envlist =
     pypy
 
 [testenv]
-commands = py.test
+commands = py.test {posargs:}
 deps =
     pytest
     pytest-cov


scls19fr:
Most of builds was broken. CI triggered again.

scls19fr:
Builds still failing

ppkt:
I believe it's caused by invoking tests by `python setup.py test` in Travis CI. Running them just by running `tox` works fine. Furthermore, it seems that using tox in setup.py is discouraged: https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command

tomschr:
Hmn, this is strange. Seems there is a difference between the run with `python setup.py test` and `tox`.

I've try a different attempt now

tomschr:
@ppkt Ha, we almost posted at the same time with the same idea. :laughing: 

I've replaced the call of `python setup.py test` with `tox -v`. Seems to work. :wink:

tomschr:
I've also added a second line with caching pip dependencies. I think this is useful and described in https://docs.travis-ci.com/user/caching
The cache is stored on Travis in `$HOME/.cache/pip`.

tomschr:
Oh, just another stupid comment before I forget: Please adjust the commit message when you squash the commits. I think it deviated a bit from the original idea.

I would suggest to use something like this (adjust it to your needs):

```
Fix #146: improve tox.ini/.travis.yml

* Changes in tox.ini:
  * add posargs for pytest. This allows to run only a single test function.
  * Rename py.test -> pytest as the old spelling py.test will be deprecated in the future.

* Changes in .travis.yml:
  * Add cache for pip to speed up things a little bit
  * Add tox as a installation requirement
  * Replace setup.py with tox as it discouraged in https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command
```

scls19fr:
Thanks again @tomschr for your contribution

PR #149, opened 2019-10-02
Fix #148: Remove & replace python setup.py test

### LINKED ISSUE #148: Doc: Remove recommendation of "setup.py test" in Contribute Section
## Situation

According to https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command, we shouldn't encourage our users to run `python setup.py test`.

Currently, there is this line in our documentation.

## Proposed Solution

Recommend to use `tox` and remove  `python setup.py test`.

tomschr:
This PR fixes #148 and contains:

As tox discouraged using "python setup.py test", we shouldn't recommend that in our documentation. Replacing it with `tox`.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -49,17 +49,17 @@ We recommend to use the following workflow if you would like to contribute:
 4. Work on your branch. Commit your work. Don't forget to write test cases
    for your new feature.
 
-5. Run the test suite. You have the following options:
+5. Run the test suite. You can decide to run the complete test suite or
+   only part of it:
 
-   * To run a complete test use the ``setup.py`` script (shown for Python 3)::
+   * To run all tests, use::
 
-     $ python3 setup.py test
+     $ tox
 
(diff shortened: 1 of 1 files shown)


tomschr:
@ppkt 
> LGTM :) Would you like to also update `setup.py` and remove invoking tox (and use pytest instead)?

Just to clarify: you mean, changing the `Tox` class in `setup.py` and replacing the `tox` command with `pytest`, right?

ppkt:
@tomschr yes, I think `python setup.py test` command should run tests for current Python instance. However, it looks like there is a discussion about removing `test` command at all: https://github.com/pypa/setuptools/issues/1684
Maybe let's leave it as it is for now :)

tomschr:
@ppkt 
> Maybe let's leave it as it is for now :)

Ok, I'm fine with that. :wink: :+1:

PR #155, opened 2019-10-03
Fix #154: Improve Description on PyPI

### LINKED ISSUE #154: Improve Description on PyPI
## Situation

The project description on https://pypi.org/project/semver/ is a bit short. Although there is a link to the documentation, people could miss it and be mislead that the project is "undocumented", in a alpha state etc.

## Proposed Solution

When developers search for something, we should support them with at least some small hints. I consider this as cheap and easy "marketing the project" in the sense of attracting people instead of repelling them. Let's make our project shine! :wink: 

Probably it's enough to insert some small information into the `README.rst`. The information could contain a small "session" how to work with it:

* Import the module
* Parse a version string
* Access the different parts of a version

Probably that would be enough to get the attention of the reader. For details, we can refer to our documentation.

tomschr:
* Improve README and describe a short session with parsing a version string and access the different parts of a version.
* Include README (=Quickstart) as part of the documentation.
* Add missing docstrings for properties

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -87,7 +87,7 @@ todo_include_todos = False
 # a list of builtin themes.
 #
 # html_theme = 'alabaster'
-html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -87,7 +87,7 @@ todo_include_todos = False
 # a list of builtin themes.
 #
 # html_theme = 'alabaster'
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -87,7 +87,7 @@
 # a list of builtin themes.
 #
 # html_theme = 'alabaster'
-html_theme = 'sphinx_rtd_theme'
+html_theme = 'alabaster'
```
scls19fr: Why changing theme?
If you want to change theme why not commenting

    html_theme = 'sphinx_rtd_theme

and commenting out

html_theme = 'alabaster'

tomschr: Uups, that was an accident and shouldn't be included into this commit. Will change it.

PR #156, opened 2019-10-05
Fix 144: Add replace function

### LINKED ISSUE #144: Extend VersionInfo class with replace method
## Situation

Let's assume the following use case: Someone wants to create a new version from an existing version, but need to change parts individually. For example, a developer wants to change only the minor part to another value. Currently, you need a "detour":

```python
import semver
v1 = semver.VersionInfo(major=3, minor=4, patch=5)
d = semver.parse(str(v))
d['minor'] = 10
v2 = semver.VersionInfo(**d)
```

(Of course there are probably other ways to get there, but I picked just this one).

IMHO, this is rather unpythonic.

## Proposed Solution

Couldn't we add an additional method `VersionInfo.replace`? With this in mind, it would be easy to update the use case above:

```python
import semver
v1 = semver.VersionInfo(major=3, minor=4, patch=5)
v2 = v1.replace(minor=10)
```

I know it's probably not a very common use case, but it could help to provide a consistent API.

Some properties about this new method:

* Accepts the usual keys  (major, minor, etc).
* It does not change the instance, it creates a new one

Questions:

* Is this a good idea?
* Should we add it into the `VersionInfo` class? 
* Or would it be a better fit if we create a module function `replace` with the signature `semver.replace(version, **kwargs)`?
* Should we raise an exception if we pass unknown keys to `replace`? Or ignore them silently?
* Any other comments?

tomschr:
This PR fixes #144 and contains:

* `semver.replace(version: str, **parts: dict)`
* `semver.VersionInfo.replace(**parts: dict)`
* Extend the usage documentation
* ~~Adapt `format_version()` function. This function accepts strings in the major, minor, and patch parts now. It was necessary as it makes the implementation of `semver.replace()` a bit easier.~~
* Both `replace()` functions raises a `KeyError` exception if you pass invalid keys. This makes the API consistent. 
  My original implementation just delegate it to the respective function; however, that leads to different exceptions raised (`TypeError` or `ValueError`).

Both do almost the same: it replaces parts of a version string or a `VersionInfo` object and creates a new object (with the changed parts).

**Update 2019-10-12:**

The `format_version()` function does not need to be adapted anymore.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,8 @@ Features
 * :pr:`125`: Created :file:`setup.cfg` for pytest and tox
 * :gh:`126` (:pr:`127`): Added target for documentation in :file:`tox.ini`
 * :gh:`142` (:pr:`143`): Improved usage section
+* :gh:`144` (:pr:`156`): Added :func:`semver.replace` and :func:`semver.VersionInfo.replace`
+  functions
 * :gh:`145` (:pr:`146`): Added posargs in :file:`tox.ini`
 * :pr:`157`: Introduce :file:`conftest.py` to improve doctests
 * :pr:`165`: Improved code coverage
--- a/semver.py
+++ b/semver.py
@@ -292,6 +292,28 @@ prerelease='pre.2', build='build.4')
         """
         return parse_version_info(version)
 
+    def replace(self, **parts):
+        """Replace one or more parts of a version and return a new
+          :class:`semver.VersionInfo` object, but leave self untouched
+
+        :param dict parts: the parts to be updated. Valid keys are:
+          ``major``, ``minor``, ``patch``, ``prerelease``, or ``build``
+        :return: the new :class:`semver.VersionInfo` object with the changed
+          parts
+        :raises: TypeError, if ``parts`` contains invalid keys
+        """
+        version = self._asdict()
+        version.update(parts)
+        try:
+            return VersionInfo(**version)
+        except TypeError:
+            unknownkeys = set(parts) - set(self._asdict())
+            error = ("replace() got %d unexpected keyword "
+                     "argument(s): %s" % (len(unknownkeys),
+                                          ", ".join(unknownkeys))
+                     )
+            raise TypeError(error)
+
 
 def _to_dict(obj):
     if isinstance(obj, VersionInfo):
@@ -702,6 +724,24 @@ def main(cliargs=None):
     except (ValueError, TypeError) as err:
         print("ERROR", err, file=sys.stderr)
         return 2
+
+
+def replace(version, **parts):
+    """Replace one or more parts of a version and return the new string
+
+    :param str version: the version string to replace
+    :param dict parts: the parts to be updated. Valid keys are:
+      ``major``, ``minor``, ``patch``, ``prerelease``, or ``build``
+    :return: the replaced version string
+    :raises: TypeError, if ``parts`` contains invalid keys
+    :rtype: str
+
+    >>> import semver
+    >>> semver.replace("1.2.3", major=2, patch=10)
+    '2.2.10'
+    """
+    version = parse_version_info(version)
+    return str(version.replace(**parts))
 
 
 if __name__ == "__main__":
--- a/test_semver.py
+++ b/test_semver.py
@@ -18,14 +18,15 @@ from semver import (VersionInfo,
                     parse,
                     parse_version_info,
                     process,
+                    replace,
                     )
 
 SEMVERFUNCS = [
     compare, createparser,
     bump_build, bump_major, bump_minor, bump_patch, bump_prerelease,
     finalize_version, format_version,
-    match, max_ver, min_ver, parse, process,
-]
+    match, max_ver, min_ver, parse, process, replace,
+    ]
 
 
 @pytest.mark.parametrize("string,expected", [
@@ -655,3 +656,45 @@ def test_should_process_raise_error(capsys):
     assert rc != 0
     captured = capsys.readouterr()
     assert captured.err.startswith("ERROR")
+
+
+@pytest.mark.parametrize("version,parts,expected", [
+    ("3.4.5", dict(major=2), '2.4.5'),
+    ("3.4.5", dict(major="2"), '2.4.5'),
+    ("3.4.5", dict(major=2, minor=5), '2.5.5'),
+    ("3.4.5", dict(minor=2), '3.2.5'),
+    ("3.4.5", dict(major=2, minor=5, patch=10), '2.5.10'),
+    ("3.4.5", dict(major=2, minor=5, patch=10, prerelease="rc1"),
+     '2.5.10-rc1'),
+    ("3.4.5", dict(major=2, minor=5, patch=10, prerelease="rc1", build="b1"),
+     '2.5.10-rc1+b1'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_replace_method_replaces_requested_parts(version, parts, expected):
+    assert replace(version, **parts) == expected
+
+
+def test_replace_raises_TypeError_for_invalid_keyword_arg():
+    with pytest.raises(TypeError, match=r"replace\(\).*unknown.*"):
+        assert replace("1.2.3", unknown="should_raise")
+
+
+@pytest.mark.parametrize("version,parts,expected", [
+    ("3.4.5", dict(major=2, minor=5), '2.5.5'),
+    ("3.4.5", dict(major=2, minor=5, patch=10), '2.5.10'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_return_versioninfo_with_replaced_parts(version,
+                                                       parts,
+                                                       expected):
+    assert VersionInfo.parse(version).replace(**parts) == \
+           VersionInfo.parse(expected)
+
+
+def test_replace_raises_ValueError_for_non_numeric_values():
+    with pytest.raises(ValueError):
+        VersionInfo.parse("1.2.3").replace(major="x")
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -118,6 +118,37 @@ In case you need the different parts of a version stepwise, iterate over the :cl
     [3, 4, 5, 'pre.2', 'build.4']
 
 
+Replacing Parts of a Version
+----------------------------
+
+If you want to replace different parts of a version, but leave other parts
+unmodified, use one of the functions :func:`semver.replace` or
+:func:`semver.VersionInfo.replace`:
+
+* From a version string::
+
+   >>> semver.replace("1.4.5-pre.1+build.6", major=2)
+   '2.4.5-pre.1+build.6'
+
+* From a :class:`semver.VersionInfo` instance::
+
+   >>> version = semver.VersionInfo.parse("1.4.5-pre.1+build.6")
+   >>> version.replace(major=2, minor=2)
+   VersionInfo(major=2, minor=2, patch=5, prerelease='pre.1', build='build.6')
+
+If you pass invalid keys you get an exception::
+
+   >>> semver.replace("1.2.3", invalidkey=2)
+   Traceback (most recent call last)
+   ...
+   TypeError: replace() got 1 unexpected keyword argument(s): invalidkey
+   >>> version = semver.VersionInfo.parse("1.4.5-pre.1+build.6")
+   >>> version.replace(invalidkey=2)
+   Traceback (most recent call last)
+   ...
+   TypeError: replace() got 1 unexpected keyword argument(s): invalidkey
+
+
 .. _sec.convert.versions:
 
 Converting Different Version Types


scls19fr:
@kxepal @k-bx @Flamefire what is your opinion about this PR?

semver.py
```diff
@@ -288,6 +288,29 @@ def parse(version):
...
+        d1.update(parts)
+        try:
+            return VersionInfo(**d1)
+        except ValueError:
+            parts.pop('major', None)
+            parts.pop('minor', None)
+            parts.pop('patch', None)
+            parts.pop('prerelease', None)
+            parts.pop('build', None)
+            keys = ", ".join([k for k in parts.keys()])
```
Flamefire: Wouldn't a set-diff be more expressive? Not sure if this works as-is but I think: `parts.keys() - d1.keys()`

Oh and I'd use a better name than `d1`

semver.py
```diff
@@ -493,8 +519,10 @@ def format_version(major, minor, patch, prerelease=None, build=None):
     >>> import semver
     >>> semver.format_version(3, 4, 5, 'pre.2', 'build.4')
     '3.4.5-pre.2+build.4'
+    >>> semver.format_version("3", 4, "5", 'pre.2', 'build.4')
+    '3.4.5-pre.2+build.4'
     """
-    version = "%d.%d.%d" % (major, minor, patch)
+    version = "%d.%d.%d" % (int(major), int(minor), int(patch))
```
Flamefire: maybe just `'{}.{}.{}'.format(major, minor, patch)`? No need to convert str->int->str
Alternative: Create a Version object (constructor handles conversion now) and do the formatting based on that object.

semver.py
```diff
@@ -615,6 +643,44 @@ def finalize_version(version):
...
+      ``major``, ``minor``, ``patch``, ``prerelease``, or ``build``
+    :return: the replaced version string
+    :raises: KeyError, when parts does not contain valid keys
+    :rtype: str
+
+    >>> import semver
+    >>> semver.replace("1.2.3", major=2, patch=10)
+    '2.2.10'
+    """
+    result = {}
```
Flamefire: Why isn't the existing implementation reused? `return str(parse(version).replace(**parts))`

I would strongly oppose having 2 implementations of the same thing.

test_semver.py
```diff
@@ -580,3 +584,44 @@ def test_should_be_able_to_use_integers_as_prerelease_build():
...
+    ("3.4.5", dict(major=2, minor=5, patch=10), '2.5.10'),
+    ("3.4.5", dict(major=2, minor=5, patch=10, prerelease="rc1"),
+     '2.5.10-rc1'),
+    ("3.4.5", dict(major=2, minor=5, patch=10, prerelease="rc1", build="b1"),
+     '2.5.10-rc1+b1'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_replace(version, parts, expected):
```
Flamefire: Better name please. The other names read like a sentence (part)

test_semver.py
```diff
@@ -580,3 +584,44 @@ def test_should_be_able_to_use_integers_as_prerelease_build():
...
+])
+def test_should_versioninfo_replace(version, parts, expected):
+    v1 = VersionInfo.parse(version)
+    assert v1.replace(**parts) == VersionInfo.parse(expected)
+
+
+def test_should_raise_value_error_for_versioninfo_replace():
+    with pytest.raises(KeyError):
+        v1 = VersionInfo.parse("1.2.3")
+        v1.replace(major="x")
```
Flamefire: A key error is wrong here. It should raise a ValueError. Good test case for that (ValueError instead of KeyError) though!

test_semver.py
```diff
@@ -580,3 +584,44 @@ def test_should_be_able_to_use_integers_as_prerelease_build():
...
+
+
+@pytest.mark.parametrize("version,parts,expected", [
+    ("3.4.5", dict(major=2, minor=5), '2.5.5'),
+    ("3.4.5", dict(major=2, minor=5, patch=10), '2.5.10'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_versioninfo_replace(version, parts, expected):
```
Flamefire: Not sure what others think, but I'd fuse the 2 test cases to reuse the parameters.   
This makes it clear that `replace(version, **parts) ==  VersionInfo.parse(version).replace(**parts) == expected` (modulo string conversion)

test_semver.py
```diff
@@ -580,3 +584,44 @@ def test_should_be_able_to_use_integers_as_prerelease_build():
...
+     '2.5.10-rc1+b1'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_replace(version, parts, expected):
+    assert replace(version, **parts) == expected
+
+
+def test_should_raise_typeerror_for_invalid_replace():
```
Flamefire: Wrong name (typeerror?)

semver.py
tomschr: > Wouldn't a set-diff be more expressive?

Good idea! See below.

> Not sure if this works as-is but I think: `parts.keys() - d1.keys()`

With sets, this won't be necessary.

> Oh and I'd use a better name than d1

Yes. :+1:  

Hope this is better:

```python
version = self._asdict()
version.update(parts)
try:
    return VersionInfo(**version)
except ValueError:
    parts.pop('major', None)
    parts.pop('minor', None)
    parts.pop('patch', None)
    parts.pop('prerelease', None)
    parts.pop('build', None)
    keys = ", ".join(set(parts) - set(version))
    raise KeyError("Invalid key(s): %s" % keys)
```

semver.py
tomschr: > maybe just '{}.{}.{}'.format(major, minor, patch)? No need to convert str->int->str

That's a good idea! :+1: 

> Alternative: Create a Version object (constructor handles conversion now) and do the formatting based on that object.

That would change the implementation of `format_version` which IMHO is something for another PR.

semver.py
Flamefire: It still suffers from a wrong conversion of ValueError->KeyError. I'd check for keyerror first (via set diff) and remove the try completely

semver.py
Flamefire: Well you added the support for string arguments in this PR. Then maybe extract that to another PR too and implement it "properly"? This avoids changing this back-and-forth

test_semver.py
tomschr: Actually that was on purpose to be consistent. My first implementation was different: Depending on which function you call (`semver.replace` or semver.VersionInfo.replace`) you would get different exceptions (ValueError vs. TypeError, if I remember correctly).

Unfortunately, the `bump_*` functions are inconsistent what they raise. So...

semver.py
tomschr: > Why isn't the existing implementation reused? `return str(parse(version).replace(**parts))`

Sorry, either I miss something, or this doesn't work.

The expression `parse(version)` returns a `dict` and that doesn't have a member function `.replace`. :wink: You probably mean something like:

```python
parse(version).update(**parts)
```

Unfortunately, even this won't work. The function `.update` updates the `dict` _in place_, returns `None`. What was your idea?

I agree though, we shouldn't have two implementations for the same thing. My idea was to check for invalid keys for the `parts` part. If you have a clever idea, I will be happy to include it.

Flamefire: Essentially my idea would be: Create a VersionInfo object, operate on that, convert to output (either a VersionInfo object or a string, as in this case). So: `str(parse_version_info).replace(**parts))`

I don't really understand the design decision that the shortest function `parse` returns a dict rather than a VersionInfo object but I guess this cannot be changed now. I'd always create a proper object first and then do manipulations on that.

This would make bump_version etc. also simpler.

semver.py
tomschr: Well, the `semver.replace(version, **parts)` delegates the formatting to the `format_version` function. If we want to allow not only integers, but also strings in the parts argument,  `format_version` needs to accept also strings.

It's really this tiny change to make str -> int in the above line 525. It wasn't my intention to change the whole implementation of that function.

If I move that change into another PR, this PR depends on the other. Which would complicate things a bit, do you agree? :wink:

Flamefire: If `semver.replace` uses `VersionInfo.replace` then you don't need that change as the ctor will make an int out of the str in `parts`. Problem solved ;)

test_semver.py
Flamefire: I'd rather be correct for new functionality and fix the old later. This ought to be a ValueError/TypeError as even the test name states

Why are the `bump_*` function inconsistent? They only take a version so only error is a ValueError. Using "x" for a version (part) is a ValueError. So KeyError is wrong here. Or am I missing anything?

tomschr: > Why are the bump_* function inconsistent?

Hmn... I can't reproduce it anymore. Probably it was due to that we coerced the major, minor, and patch arguments to integers which changed the exceptions too... :thinking: 

Anyway, yes, I agree, in that case it would be better to check for a ValueError exception. That would avoid the try...except handling after line 302; it would be unnecessary.

semver.py
tomschr: I think, I took your idea and reduced it to the bare minimum:

```python
def replace(version, **parts):
    """Replace one or more parts of a version and return the new string
    [...]
    """
    version = parse_version_info(version)
    return str(version.replace(**parts))
```

However, when I try to pass invalid keys, I got a `TypeError`:

```python
>>> semver.replace("1.2.3", major=2, patch=10, answer=42)
Traceback (most recent call last)
...
TypeError: __init__() got an unexpected keyword argument 'foo'
```

Looks a bit ugly, doesn't it? :wink:

Flamefire: Yes, this was what I had in mind.

Oh ok, so a wrong keyword is a TypeError not a KeyError... Well if you want it to be a KeyError and/or make the message better you need to adjust the replace function as outlined in https://github.com/k-bx/python-semver/pull/156/files#r332368167

However as Python is raising a TypeError in such cases I'd vote for a typeerror too and make the error message match: `TypeError(replace() got an unexpected keyword argument 'answer')`

test_semver.py
tomschr: I'm bad at inventing good names. :wink: Maybe something like `test_should_compare_version_and_replace_parts`?

Flamefire: `test_replace_method_replaces_request_parts`? This is what is done here ;)

test_semver.py
tomschr: Yes, I've used now `test_should_raise_keyerror_for_invalid_key_in_replace`

test_semver.py
tomschr: I've corrected this.

semver.py
tomschr: I used your last suggestion and raised a `TypeError`.

tomschr:
@Flamefire Sorry for the delay, but I was on a business trip.

Alexander, I've followed your suggestions as much as possible. Hope I didn't miss anything. I've tried to "delegate" things to other functions to avoid implementing the same. (for example, the `format_version` function isn't used anymore).

Feel free to send me your ideas! Much appreciated. :+1:

scls19fr:
Similar to https://github.com/k-bx/python-semver/pull/164#issuecomment-541443311
Updating [CHANGELOG.rst](https://github.com/k-bx/python-semver/blob/master/CHANGELOG.rst) will be nice.

CHANGELOG.rst
```diff
@@ -22,6 +22,7 @@ Features
 * :pr:`125`: Created :file:`setup.cfg` for pytest and tox
 * :gh:`126` (:pr:`127`): Added target for documentation in :file:`tox.ini`
 * :gh:`142` (:pr:`143`): Improved usage section
+* :gh:`144` (:pr:`156`): Added replace function
```
scls19fr: Maybe `replace` for syntax highlighting

tomschr: I have rephrased the line to this:

```rst
* :gh:`144` (:pr:`156`): Added :func:`semver.replace` and :func:`semver.VersionInfo.replace`
  functions
```

This has the advantage, that the reader can click on the referenced functions.

scls19fr: Very good idea

scls19fr:
Could you fix conflicting files?

tomschr:
> Could you fix conflicting files?

Uups! Sorry, will take care of that.

scls19fr:
@ppkt could you review?

test_semver.py
```diff
@@ -655,3 +656,43 @@ def test_should_process_raise_error(capsys):
...
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_versioninfo_replace(version, parts, expected):
+    assert VersionInfo.parse(version).replace(**parts) == \
+           VersionInfo.parse(expected)
+
+
+def test_should_raise_value_error_for_versioninfo_replace():
```
Flamefire: "replace_should_raise_ValueError_for_unknown_keyword_args"

test_semver.py
```diff
@@ -655,3 +656,43 @@ def test_should_process_raise_error(capsys):
...
+
+
+@pytest.mark.parametrize("version,parts,expected", [
+    ("3.4.5", dict(major=2, minor=5), '2.5.5'),
+    ("3.4.5", dict(major=2, minor=5, patch=10), '2.5.10'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_versioninfo_replace(version, parts, expected):
```
Flamefire: "versioninfo_replace_should_return_versioninfo_with_replaced_parts"

test_semver.py
```diff
@@ -655,3 +656,43 @@ def test_should_process_raise_error(capsys):
...
+    ("3.4.5", dict(major=2, minor=5, patch=10), '2.5.10'),
+    ("3.4.5", dict(major=2, minor=5, patch=10, prerelease="rc1"),
+     '2.5.10-rc1'),
+    ("3.4.5", dict(major=2, minor=5, patch=10, prerelease="rc1", build="b1"),
+     '2.5.10-rc1+b1'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_replace_method_replaces_request_parts(version, parts, expected):
```
Flamefire: request**ed*_parts

test_semver.py
tomschr: @Flamefire 
Thanks, I don't mind changing that, but that makes the line **very** long. :wink: Additionally, you get a Flake8 E501 error ("line too long"). Could we have a shorter name?

By the way, have you worked with [GitHub Suggestions (step 6)](https://help.github.com/en/articles/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request)? It makes it a bit easier to work with feedback. You only click the "+/-" icon (the one between "Preview" and the "AA") and change the line to what you would like to have. It gives you a diff-like view.

test_semver.py
tomschr: Done, corrected it to `test_replace_should_raise_ValueError_for_unknown_keyword_args`.

test_semver.py
tomschr: Done

test_semver.py
```diff
@@ -655,3 +656,45 @@ def test_should_process_raise_error(capsys):
...
+     '2.5.10-rc1+b1'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_replace_method_replaces_requested_parts(version, parts, expected):
+    assert replace(version, **parts) == expected
+
+
+def test_should_raise_keyerror_for_invalid_key_in_replace():
```
Flamefire:
```suggestion
def test_replace_raises_ValueError_for_invalid_keyword_arg():
```

test_semver.py
```diff
@@ -655,3 +656,45 @@ def test_should_process_raise_error(capsys):
...
+    with pytest.raises(TypeError):
```
Flamefire: Why `TypeError`? Shouldn't this be a ValueError as in https://github.com/k-bx/python-semver/pull/156/files#diff-a9f8e0a40655687a794b987c4b31ff9fR699?

test_semver.py
tomschr: Ok, done. :+1:

test_semver.py
tomschr: Well, it's because if you pass an unknown key to `semver.VersionInfo`, the `__init__` [raises a TypeError](https://github.com/tomschr/python-semver/blob/feature/144-replace/semver.py#L309), for example:

```python
>>> import semver
>>> semver.VersionInfo(major=1, foo=42)
Traceback (most recent call last)
 ...
TypeError: __init__() got an unexpected keyword argument 'foo'
```

The current implementation catches the TypeError and improves the error message:

```python
>>> semver.VersionInfo(major=1).replace(foo=42) 
Traceback (most recent call last)
 ...
TypeError: replace() got 1 unexpected keyword argument(s): foo
```

Of course, I can rewrite the exception to KeyError or ValueError if you like. However, that means, we need to adapt that in [`semver.replace`](https://github.com/tomschr/python-semver/blob/feature/144-replace/semver.py#L729)  as well.

Flamefire: My point is: The free function and the member function should do the same thing. Why does one raise a ValueError while the other raises a TypeError in the same situation? As Python itself raises a TypeError on unexpected keyword arg then the improved error should be a TypeError too. So it seems https://github.com/k-bx/python-semver/pull/156/files#diff-a9f8e0a40655687a794b987c4b31ff9fR699 need to be adapted.

I've just seen: They test different things. So it is the name that is wrong.

test_semver.py
```diff
@@ -655,3 +656,45 @@ def test_should_process_raise_error(capsys):
...
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_should_return_versioninfo_with_replaced_parts(version,
+                                                       parts,
+                                                       expected):
+    assert VersionInfo.parse(version).replace(**parts) == \
+           VersionInfo.parse(expected)
+
+
+def test_replace_should_raise_ValueError_for_unknown_keyword_args():
```
Flamefire:
```suggestion
def test_replace_raises_ValueError_for_non_numeric_values():
```

test_semver.py
Flamefire: Made a rename suggestion. Now it should be all good :)

test_semver.py
tomschr: Ok, renamed :wink: in commit 4df2176

test_semver.py
```diff
@@ -655,3 +656,45 @@ def test_should_process_raise_error(capsys):
...
+     '2.5.10-rc1+b1'),
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_replace_method_replaces_requested_parts(version, parts, expected):
+    assert replace(version, **parts) == expected
+
+
+def test_replace_raises_ValueError_for_invalid_keyword_arg():
```
Flamefire: Sorry, missed that:

```suggestion
def test_replace_raises_TypeError_for_invalid_keyword_arg():
```

test_semver.py
tomschr: > My point is: The free function and the member function should do the same thing.

I agree. :+1: 

> So it seems https://github.com/k-bx/python-semver/pull/156/files#diff-a9f8e0a40655687a794b987c4b31ff9fR699 need to be adapted.

Hmn, ... :thinking: ... after looking at both implementations, it seems, it's easier to change the (already existing) TypeError into a ValueError in [line 315](https://github.com/tomschr/python-semver/blob/feature/144-replace/semver.py#L315). 

@Flamefire 
Would that be an option? Or should I better stay with TypeError (and adapting `semver.replace` as well)?

Thanks for all your valuable feedback. Highly appreciated! :+1:

Flamefire: > after looking at both implementations, it seems, it's easier to change the (already existing) TypeError into a ValueError i

No. I noticed my mistake while typing that message, see the last sentence: One tests an invalid keyword arg which should result in a TypeError and the other tests an invalid value for a valid keyword arg which should result in a ValueError.  I missed that little difference and just assumed the functions would test the same thing.

So just adapt the name of the test function so they match the behavior (see suggestion), I think everything else is fine, so this should be the last change.

Still one possible improvement: You could test the error message too (that it contains `replace()` and the wrong keyword arg) if you wanted. I don't consider that important, but I've worked with people who do.

tomschr: > So just adapt the name of the test function so they match the behavior (see suggestion), I think everything else is fine, so this should be the last change.

Ok, thanks, I did that now.

> Still one possible improvement: You could test the error message too (that it contains replace() and the wrong keyword arg) if you wanted. I don't consider that important, but I've worked with people who do.

I thought of that too, but was a bit unsure if I should implement it. But you are right, it could be an improvement. Will do so. :wink:

test_semver.py
tomschr: Implemented in commit 3a758a2

test_semver.py
tomschr: Implemented in commit 3a758a2

test_semver.py
```diff
@@ -655,3 +656,52 @@ def test_should_process_raise_error(capsys):
...
+    ("3.4.5-alpha.1.2", dict(major=2), '2.4.5-alpha.1.2'),
+    ("3.4.5-alpha.1.2", dict(build="x1"), '3.4.5-alpha.1.2+x1'),
+    ("3.4.5+build1", dict(major=2), '2.4.5+build1'),
+])
+def test_replace_method_replaces_requested_parts(version, parts, expected):
+    assert replace(version, **parts) == expected
+
+
+def test_replace_raises_TypeError_for_invalid_keyword_arg():
+    with pytest.raises(TypeError):
```
Flamefire: You can use the following to avoid having a 2nd test case for the same:

```suggestion
    with pytest.raises(TypeError, match=r"replace\(\).*unknown.*"):
```

See http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

tomschr: Brilliant! :smile: I searched for a better way, now as you mention it. :+1: It's now implemented in commit 5049b83

tomschr:
Thanks Alexander for your great help! :+1: Much appreciated, I learned a lot. :heart: 

@scls19fr, @k-bx, @ppkt Any further comments/wishes/ideas from your side? :wink:

PR #158, opened 2019-10-06
Doc: Clarify Contributing section

tomschr:
This PR does not belong to a specific GH issue. It's a PR with doc releated changes only:

* Structure the text with more subsection
* Introduce a separate section about "Running the Test Suite"
  This gives an overview, recommends flake8, and how to run
  the complete test suite or only a specific test function
* Introduce a separate section about "Documenting semver"
  Gives contributors some helpful advise about docstrings and the documentation
* Use references (`:ref:`) to other section
  This helps to reduce the "big" procedure and make it more readable.
* Change wording, correct phrases, clarify text
  As well as in Python's Zen "Simple is better than complex", this applies to documentation too.
* setup.cfg: remove `-q` option
  Removing it does NOT change the default output. However, this was needed to allow to pass `-v` to pytest (print the names of all test functions). This is covered in the documentation now.


scls19fr:
Thanks @tomschr for this contribution which greatly improve doc.

PR #164, opened 2019-10-13
Fix #59: Implement command line interface

### LINKED ISSUE #59: Command Line Interface
I've been working on a continuous integration pipeline that needs to perform very basic operations on an applications current version. The application's pipeline is currently a series of Shell commands strung together and while it would be possible to rework much of the pipeline as a Python script that uses python-semver as-is, extending python-semver with a command line interface ended up being a much lower level of effort.

While the interface I wrote is far from comprehensive, I was wondering if a CLI was something you'd be interested in having as part of the python-semver core package.

Here is a Gist of what I'd written for my immediate needs: https://gist.github.com/mattrohland/43b47bd079bd3d47adf674cbdbd970c3

tomschr:
This PR fixes #59 and contains the following changes:

* Extend setup.py with entry_point key and point to `semver.main`. The script is named `semver`

* Introduce 3 new functions:
  * `createparser`: creates and returns an `argparse.ArgumentParser` instance
  * `process`: process the CLI arguments and call the requested actions
  * `main`: entry point for the application script

* Add test cases:
  * sort import lines of semver functions/class with isort tool
  * sort list of SEMVERFUNCS variable

* Extend documentation:
  * Add `sphinx-argparse` as a doc requirement to document `semver` script (subcommands and options)
  * Include new `cli.rst` file which (self)documents the arguments of the semver script using the `sphinx-argparse` plugin
  * Extend extensions variable in `conf.py` to be able to use the `sphinx-argparse` module

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Version 2.9.0 (WIP)
 Features
 --------
 
+* :gh:`59` (:pr:`164`): Implemented a command line interface
 * :gh:`85` (:pr:`147`, :pr:`154`): Improved contribution section
 * :gh:`104` (:pr:`125`): Added iterator to :func:`semver.VersionInfo`
 * :gh:`112`, :gh:`113`: Added Python 3.7 support
@@ -24,6 +25,7 @@ Features
 * :gh:`142` (:pr:`143`): Improved usage section
 * :gh:`145` (:pr:`146`): Added posargs in :file:`tox.ini`
 * :pr:`157`: Introduce :file:`conftest.py` to improve doctests
+
 
 Bug Fixes
 ---------
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1,0 +1,45 @@
+CLI
+===
+
+The library provides also a command line interface. This allows to include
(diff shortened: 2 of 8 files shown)


scls19fr:
Thanks @tomschr for this
As we are close to a new release, I think we should always have CHANGELOG updated also.
Could you update it accordingly in all current PR?

tomschr:
@scls19fr 
> As we are close to a new release, I think we should always have CHANGELOG updated also.
> Could you update it accordingly in all currents PR?

Thanks for the hint! I agree, I should have better updated the CHANGELOG too. Will do that. :+1:

tomschr:
@scls19fr I've realized that my implementation has an issue. The current name of the script is `semver`. However, npm-semver defines an identical name.

To avoid any conflicts for packagers and users, I fear we should use a different name. Maybe `semver.py`, `pysemver`, or something similar? What do you prefer?

Thanks for your ideas! :+1:

scls19fr:
I think `pysemver` is very explicit about what it is (and it's a filename without extension)

scls19fr:
Thanks @tomschr

PR #168, opened 2019-10-13
Fix #167: Introduce SEMVER_SPEC_VERSION

### LINKED ISSUE #167: Introduce SEMVER_SPEC_VERSION variable for Semantical Version API
## Situation

After studying the source code of [npm-semver](https://github.com/npm/node-semver/), I've found that `npm-semver` defines a [`SEMVER_SPEC_VERSION`](https://github.com/npm/node-semver/blob/master/semver.js#L20) variable.

At the moment, semver doesn't have a similar variable. I think, this could be useful for this project as well.

## Proposed Solution

I would propose to introduce a global variable `SEMVER_SPEC_VERSION` with the value:

```python
SEMVER_SPEC_VERSION = "2.0.0"
```

The variable defines, which API version of Semantical Version is implemented by semver.

It's a very simple change and shouldn't introduce any problems.

tomschr:
This PR fixes #167 and implements the global SEMVER_SPEC_VERSION constant.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,7 @@ Features
 * :gh:`142` (:pr:`143`): Improved usage section
 * :gh:`145` (:pr:`146`): Added posargs in :file:`tox.ini`
 * :pr:`157`: Introduce :file:`conftest.py` to improve doctests
+* :gh:`167` (:pr:`168`): Introduced global constant :data:`SEMVER_SPEC_VERSION`
 
 Bug Fixes
 ---------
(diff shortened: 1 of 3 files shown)


scls19fr:
Similar to https://github.com/k-bx/python-semver/pull/164#issuecomment-541443311
Updating [CHANGELOG.rst](https://github.com/k-bx/python-semver/blob/master/CHANGELOG.rst) will be nice.

scls19fr:
Thanks @tomschr

PR #170, opened 2019-10-30
Release 2.9.0

### LINKED ISSUE #134: Publish a new release on PyPi (2.9.0)
Thanks to great work of @Flamefire we are now ready to publish a new release on PyPi

Here is a release procedure to follow
https://github.com/k-bx/python-semver/blob/master/release-procedure.md

but maybe we should have a discussion before about version numbering.

I don't really know if this new version should be named 2.8.2 or 2.9.0 (to respect semver)

Your opinions?

PS : I won't be available during the next coming days (until next week) so either @k-bx can publish or we can wait some days.

scls19fr:
Closes #134

### DIFF
--- a/release-procedure.md
+++ b/release-procedure.md
@@ -1,8 +1,12 @@
+#################
+Release procedure
+#################
+
 * Verify that latest build was passing https://travis-ci.org/k-bx/python-semver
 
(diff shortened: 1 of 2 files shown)


tomschr:
Thanks for all your efforts! :+1:

PR #175, opened 2019-10-31
First GitHub Action to test for test.pypi.org

tomschr:
Taken from https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/


tomschr:
Ahh, sorry, wrong repo. :-(

PR #195, opened 2019-11-15
Fix #191: Add manpage for pysemver

### LINKED ISSUE #191: Create Man Page for pysemver
## Situation

When creating packages for Linux distributions (RPM or DEB), it is good practice to have a man page for each executable.

However, we currently don't have a dedicated man page for our `pysemver` script. Although we have rudimentary descriptions in our  `cli.rst`, it doesn't follow the man page style. 

## Proposed Solution

We should provide the man page for this project. When I've wrote this  `cli.rst` file, I didn't think of creating a man page. I better should have thought about that.

As such, I would suggest to improve the existing `cli.rst` page and turn it into a "real" man page. As we probably add more features in the future, we should prepare it to help users and packages.

I would propose to change the existing `cli.rst` file and add the usual man page structure:

```
NAME
    pysemver - Helper script for Semantic Versioning

SYNOPSIS
   pysemver compare version1 version2
   pysemver bump {major,minor,patch,prerelease,build} version

DESCRIPTION
   [...]
SEE ALSO
   https://python-semver.readthedocs.io/en/latest/

AUTHOR
   [...]
```

This can easily built with `sphinx-build` and choosing a man builder. After the man page was built, it can be added to the RPM or Debian package.

tomschr:
This PR fixes #191 and contains the following changes.

* Rename `cli.rst` -> `pysemver.rst` to be consistent with the name of our little script
* Add missing `--version` in `pysemver` script
* Restructure `pysemver.rst` to reflect manpage style
* Add additional target "man" in tox.ini; it's now possible to run "tox -e man". Manpage can be found at `docs/_build/man`.
* Ignore `docs/_build/`
* Update `CHANGELOG.rst`

---

Update 2019-11-17:
Should we include the addition of the `--version` option into this this issue? Or better create a separate one?


docs/conf.py
```diff
@@ -36,7 +36,7 @@
     'sphinx.ext.intersphinx',
     'sphinx.ext.napoleon',
     'sphinx.ext.extlinks',
-    'sphinxarg.ext',
+    # 'sphinxarg.ext',
```
ppkt: I'd remove this code instead of commiting commented out

tox.ini
```diff
@@ -22,3 +22,11 @@ basepython = python3
 deps = -r{toxinidir}/docs/requirements.txt
 skip_install = true
 commands = sphinx-build {posargs:-E} -b html docs dist/docs
+
+[testenv:man]
+whitelist_externals = make
+basepython = python3
+# deps = -r{toxinidir}/docs/requirements.txt
```
ppkt: same here, I believe it was done for testing purposes :)

tomschr:
Thank you @ppkt ! Much appreciated your review. :+1: 

Actually, it's still a draft that's why I've created it as a "draft pull request". I need to check a few things, but I pushed it anyway to have a backup. :smile: 

There is one thing that I would like to ask you about *your* opinion. What do you think about the `--version` option? 

Actually it's not a real "man page only" issue. I could open a separate issue and pull request, but on the other hand it's kind of related. I could also add a small test case. So I'm thinking of to split it into two separate PRs. Good/bad? :wink:

docs/conf.py
tomschr: Yes, that is not needed anymore as I've documented the options manually.  That's one of the issues why I've opened it as "draft". :smile:

tox.ini
tomschr: Yep. Actually I thought it could make sense to add the doc and manpage dependencies in a separate file. That makes it easier to append anything and we don't need to touch the `tox.ini` file. On the other hand, currently it's just sphinx.

There is still a little inconsistency between the targets "man" and "docs". The "man" target uses `make -C docs man` to build the manpage whereas the "docs" target uses `sphinx-build`.

I'm a big fan of consistency, so IMHO this should be consolidated. It would be better if both use the make call. Not sure if I should change that here as well... :thinking:

tomschr: I've corrected your suggestion in commit 30a5264

docs/conf.py
tomschr: I've corrected your suggestion in commit 30a5264

scls19fr:
Thanks @tomschr for this man page.

ppkt:
@tomschr in _my_ opinion - it's better to have separate PR for each "feature" but in this case, change is very small and could be added as new commit (but under the same PR)

tomschr:
@ppkt Thanks Karol for your opinion. :+1: Actually, it's already done and merged, so it's too late. But I will take it into account next time.

PR #197, opened 2019-11-20
Add distribution specific install hints

### LINKED ISSUE #196: Improve Installation Section and add distribution specific hints
## Situation

_Originally posted by @tomschr in https://github.com/python-semver/python-semver/issues/191#issuecomment-553850216_

> Distributing python-semver as package for some Linux distros should be considered.
Maybe we should only consider for now Debian and Suse.

Good idea, that inspired me to another idea: why not extend the "Installing semver" section in our documentation? We could add where to get your package for your specific distribution and how to install it (via `apt-get`, `dnf`, or `zypper`).

## Proposed Solution

Improve the Installation section and add some small hints where to find the python-semver package for different distributions (and how to install it).

tomschr:
This PR fixes #196 

* Improve the installation section and add installation instructions for Arch Linux, Debian, Fedora, FreeBSD, openSUSE, and Ubuntu.
* Adapt `CHANGELOG.rst`

----

I've mostly used the installation instructions from https://pkgs.org, but I'm not sure if this is all correct (I mainly use openSUSE). So please check, if this is ok.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ Features
 * :pr:`179`: Added note about moving this project to the new python-semver organization on GitHub
 * :gh:`187` (:pr:`188`): Added logo for python-semver organization and documentation
 * :gh:`191` (:pr:`194`): Created manpage for pysemver
+* :gh:`196` (:pr:`197`): Added distribution specific installation instructions
+
 
 Bug Fixes
 ---------
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -12,3 +12,79 @@ For Python 3:
 .. code-block:: bash
 
     pip3 install semver
+
+
+.. note::
+
+   Some Linux distributions can have outdated packages.
+   These outdated packages does not contain the latest bug fixes or new features.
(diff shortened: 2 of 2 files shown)


tomschr:
Thanks Karol for your feedback!

> Looks great, thank you :)

Thanks. :blush: 

> I would add information that in some distros (ekhem, Debian) semver package is outdated and won't have latest bugfixes or features :)

Yes, that's also what I've noticed. As this affects all distributions, it has to be a general note.  Maybe something like this:

```
You will notice that some Linux distributions can have outdated packages.
These outdated packages does not contain the latest bug fixes or new features.
If you need a newer package, you have these option:

* Ask the packager to update the package.
* Update the package for your favorite distribution and submit it.
* Use a virtual environment and `pip install`.
```

I'm a bit still unsure about the wording, but this is the idea.

ppkt:
@tomschr yes, it's very clear now :)
In meantime I'll check who is responsible for semver in Ubuntu / Debian and I'll ask them about possible update.

scls19fr:
Thanks guys for your involvment in this project!

Issue #114, opened 2018-11-25
Library is partially non-compliant with semver spec


dbarella:
Breaking input: `0.0.0-0foo.1`

Expected result:

```
>>> import semver
>>> semver.parse('0.0.0-0foo.1')
{'build': None, 'major': 0, 'patch': 0, 'minor': 0, 'prerelease': '0foo.1'}
```

Actual result:

```
>>> import semver
>>> semver.parse('0.0.0-0foo.1')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/semver.py", line 65, in parse
    raise ValueError('%s is not valid SemVer string' % version)
ValueError: 0.0.0-0foo.1 is not valid SemVer string
```

Per [item 9](https://semver.org/#spec-item-9), only numeric IDs cannot contain leading zeros -- since `0foo` is not entirely numeric, it may contain a leading zero.

scls19fr:
I also understand it this way:

    Numeric identifiers MUST NOT include leading zeroes.

so

    Non numeric identifiers MAY include leading zeroes.


What is your opinion @k-bx ?

PR (with fix and test) are welcome @dbarella !

sirzooro:
Any update on this? We use git commit hash as a prerelease part, so sometimes we trigger this error too.

tomschr:
If I'm not mistaken, it is related (or a duplicate?) of #194.

sirzooro:
Yes, it is a duplicate. You can close this issue.

tomschr:
Duplicate of #194 so closing.

PR #198, opened 2019-11-20
Fix #194 and #114: Update semver regex to version from semver.org

### LINKED ISSUE #194: Semver validation error when prerelease version contains a floating number
Could someone please verify if that's intended? I can't see any restriction about this in the semver documentation.

PoC:
```
$ python3 --version
Python 3.6.8
$ pip3 list --format=legacy |grep semver
semver (2.9.0)
$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import semver
>>> semver.parse('0.1.0-20191919191919.1fe6b123')
{'major': 0, 'minor': 1, 'patch': 0, 'prerelease': '20191919191919.1fe6b123', 'build': None}
>>> semver.parse('0.1.0-20191919191919.0fe6b123')
  File "<stdin>", line 1, in <module>
  File "/home/ezlukga/.local/lib/python3.6/site-packages/semver.py", line 72, in parse
    raise ValueError('%s is not valid SemVer string' % version)
ValueError: 0.1.0-20191919191919.0fe6b123 is not valid SemVer string
>>> semver.parse('0.1.0-20191919191919.0f')
  File "<stdin>", line 1, in <module>
  File "/home/ezlukga/.local/lib/python3.6/site-packages/semver.py", line 72, in parse
    raise ValueError('%s is not valid SemVer string' % version)
ValueError: 0.1.0-20191919191919.0f is not valid SemVer string
>>> semver.parse('0.1.0-0f')
  File "<stdin>", line 1, in <module>
  File "/home/ezlukga/.local/lib/python3.6/site-packages/semver.py", line 72, in parse
    raise ValueError('%s is not valid SemVer string' % version)
ValueError: 0.1.0-0f is not valid SemVer string
```

ppkt:
I took a regex from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string , splitted it into multiple lines to match current format and renamed `buildmetadata` -> `build`. Also, I found and fix a problem with invalid Python 2.7 super call :)

### DIFF
--- a/test_semver.py
+++ b/test_semver.py
@@ -70,6 +70,30 @@ def test_fordocstrings(func):
         'prerelease': 'alpha-1',
         'build': 'build.11.e0f985a',
         }),
+    ("0.1.0-0f",
+     {
+         'major': 0,
+         'minor': 1,
+         'patch': 0,
+         'prerelease': '0f',
+         'build': None,
+     }),
+    ("0.0.0-0foo.1",
+     {
+         'major': 0,
+         'minor': 0,
+         'patch': 0,
+         'prerelease': '0foo.1',
+         'build': None,
+     }),
+    ("0.0.0-0foo.1+build.1",
+     {
+         'major': 0,
+         'minor': 0,
+         'patch': 0,
+         'prerelease': '0foo.1',
+         'build': 'build.1',
+     }),
 ])
 def test_should_parse_version(version, expected):
     result = parse(version)
--- a/semver.py
+++ b/semver.py
@@ -19,18 +19,18 @@ __maintainer_email__ = "s.celles@gmail.com"
 _REGEX = re.compile(
         r"""
         ^
-        (?P<major>(?:0|[1-9][0-9]*))
+        (?P<major>0|[1-9]\d*)
         \.
-        (?P<minor>(?:0|[1-9][0-9]*))
+        (?P<minor>0|[1-9]\d*)
         \.
-        (?P<patch>(?:0|[1-9][0-9]*))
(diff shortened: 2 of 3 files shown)


test_semver.py
```diff
@@ -70,6 +70,22 @@ def test_fordocstrings(func):
...
+         'build': None,
+     }),
+    ("0.0.0-0foo.1",
+     {
+         'major': 0,
+         'minor': 0,
+         'patch': 0,
+         'prerelease': '0foo.1',
+         'build': None,
+     }),
```
tomschr: I would add another test case with an additional prerelease:

```suggestion
    ("0.0.0-0foo.1+build.1",
     {
         'major': 0,
         'minor': 0,
         'patch': 0,
         'prerelease': '0foo.1',
         'build': 'build.1',
     }),
```

scls19fr:
Nice catch about super call with Py 2.7
I can not wait for it to die 😅

test_semver.py
ppkt: right, added :)

scls19fr:
@ppkt Could you add a note to [CHANGELOG](https://github.com/python-semver/python-semver/blob/master/CHANGELOG.rst)?

scls19fr:
Thanks @ppkt for this PR and @tomschr for reviewing.

ppkt:
@scls19fr of course, I'll do this in next PR :)

Issue #186, opened 2019-11-09
Infra: consider using GitHub Actions?


tomschr:
## Situation

Although GitHub Actions are still in public beta, it could be a helpful addition to this project in the future.

## Ideas & Questions

Actually not any solution (yet), but some questions. :wink: 

### Scope

I've opened this issue mere to collect ideas and discuss the pros and cons. It can lead to an implementation or that members decide to better reject this issue. Both would be fine for me. :wink: 

Currently, I don't see GH Actions as a replacement for the existing Web hooks for Travis or ReadTheDocs. They don't need to be replaced (although they potentially could). I see GH Actions more as an addition to simplify some specific workflows so that these workflows appear under one roof.

Additionally, if we need a specific workflow in the future, it can be usually added quite easily. Doing the same by creating a new Web hook and the respective infrastructure can be a bit more complicated.

### Potential Use Cases
GH Actions are more than just "Travis CI". To name a few use cases:

* help with creating a release (see https://github.com/actions/create-release)
* help with creating additional assets when creating a release (for example, build automatically the documentation for each release and make it available as additional download in PDF or EPUB format; see https://github.com/actions/upload-release-asset)
* runs the black code formatter on the source code to maintain code style (see https://github.com/marketplace/actions/black-code-formatter)
* deploy automatically to Pypi after a successful build
* start automatic tasks every week/month, for example, to search for dead links, unsorted imports etc. and open new issues automatically. Like a weekly/monthly health report.

Of course, it is possible to create your own GH Actions: add a new repository to your organization and refer to it in the respective action config file.

---

Could that be a useful addition to our project?

Currently, I see the biggest benefits when creating new releases. When creating a new release, this usually involves some manual tasks which can be easily forgotten. With a GH Action this could be done automatically.

## See also

1. https://github.com/features/actions
1. https://github.com/marketplace?type=actions
1. https://help.github.com/en/actions/automating-your-workflow-with-github-actions

scls19fr:
Being able to publish release directly from Github will be a nice feature.

I was very interested in GitHub Actions when GH publish it publicly but I didn't had time (or didn't spent time) on it.

All your suggestions are very interesting maybe we should first try one usecase which could have not much "side" effects so it will be possible to later tackle harder usecases.

I personally like Black (Go have quite similar command for formatting automatically code: gofmt).

Maybe this could be a quite interesting use case which could be easily tested.

tomschr:
I fully agree with you. :+1: 

[...]

PR #200, opened 2019-11-24
Extend #186: Add GitHub Action for black formatter

tomschr:
## Scope

This PR is a first start of GH Actions to call the `black` source code formatter and does basically this:

1. Triggered by a pull request event.
2. Setup Python 3.7.
3. Install black.
4. Run black and create a diff file.
5. Upload the diff as artifact; this can be downloaded and examined if needed. See my [action run in my fork](https://github.com/tomschr/python-semver/pull/2/checks?check_run_id=318204067).

For the time being, it's quite "simple". This was on purpose as I've played with other fancy stuff which didn't work as expected (probably it was my fault because this topic is quite new for me).

From a design point of view, it is built upon other actions, created by GitHub itself.

At the moment, black creates a >50KB diff of the source code of all `*.py` files. Maybe we should create another issue to (re)format the source code with black to reduce the diff for future pull requests.

## Not Implemented Ideas

I've played around in my fork to implement other more fancy tasks. However, the result were mixed. That's why I thought I start with something more simple. If wanted, we can extend it later. :wink: 

To create your own Action you either a) reuse existing Actions, or b) create your own. I've tried to start with a) first as for b) you need to know JavaScript or Docker.

I've tried the following, but didn't include it into this PR:

* Add the diff as a comment
  I've tried to add a comment and the diff to a pull request. That would be nice so the owner of the PR would see the result from the GH Action. The comment was visible, but not the diff. For some unknown reason, I couldn't pass the result from the black step into the commenting step.

  As the diff can be quite huge, it would add a lot of "noise" to the comments. One way to avoid it would be to add only the first x lines or to use the program `filterdiff`. That can extract the first 2-4 "hunks" of a diff (use `filterdiff -#3 project.diff`). 

* Add pip caching
  Adding pip caching speeds up things. I've tested it, but it didn't work reliably for me.

## Future Ideas

If we add more GH Actions to the repo in the future, it would probably interesting to create our own Actions.

As we have now a python-semver organization, it is easy to create such additional repositories under our domain. This could be source code formatting, release code management, or something else.

I think, this is probably something we should consider in the future. With our own GH Actions, we have more finer control and the Actions are separated from the semver source code.

But this is something beyond this simple PR; I've added it here to have a place for future reference.

### DIFF
--- a/.github/workflows/black-formatting.yml
+++ b/.github/workflows/black-formatting.yml
@@ -1,0 +1,49 @@
+name: Black Formatting
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Output env variables
+        run: |
+           echo "GITHUB_WORKFLOW=${GITHUB_WORKFLOW}"
+           echo "GITHUB_ACTION=$GITHUB_ACTION"
+           echo "GITHUB_ACTIONS=$GITHUB_ACTIONS"
+           echo "GITHUB_ACTOR=$GITHUB_ACTOR"
+           echo "GITHUB_REPOSITORY=$GITHUB_REPOSITORY"
+           echo "GITHUB_EVENT_NAME=$GITHUB_EVENT_NAME"
+           echo "GITHUB_EVENT_PATH=$GITHUB_EVENT_PATH"
+           echo "GITHUB_WORKSPACE=$GITHUB_WORKSPACE"
+           echo "GITHUB_SHA=$GITHUB_SHA"
+           echo "GITHUB_REF=$GITHUB_REF"
+           echo "GITHUB_HEAD_REF=$GITHUB_HEAD_REF"
+           echo "GITHUB_BASE_REF=$GITHUB_BASE_REF"
+           echo "::debug::---Start content of file $GITHUB_EVENT_PATH"
+           cat $GITHUB_EVENT_PATH
+           echo "\n"
+           echo "::debug::---end"
+
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip black
+
+      - name: Run black
+        id: black
+        run: |
+          black . > project.diff
+          echo "::set-output name=rc::$?"
+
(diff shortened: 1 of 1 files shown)


scls19fr:
LGTM

Absolutely ok for a second PR to reformat code with Black

I'm just wondering if this second PR shouldn't be merged first (letting just one ore 2 minor formatting mistakes to ensure that this Github Action is working fine)

tomschr:
@scls19fr Thanks. :+1: 

> I'm just wondering if this second PR shouldn't be merged first (letting just one ore 2 minor formatting mistakes to ensure that this Github Action is working fine)

I've tested the GH Action in my fork and it creates the diff file as artifact. This artifact is only visible inside the Action tab, not in the repo itself. Or in other words: this GH Action does never change the source itself. So I think, it would be safe to merge it. :-)

What about this attempt:
If we merge it, we could see if this works (=> see the huge diff as an artifact). In a second pull request, we could format the source code and we could observe there no diff.

scls19fr:
Ok so let's merge.

scls19fr:
Maybe we should have a mention about that in CHANGELOG.
Sorry it's merged... but maybe you can it in second PR about Black formatting of current code.

tomschr:
Regarding the CHANGELOG: I thought about that too, but it's about the infrastructure, not the source code.

All fine! :+1: :-)

scls19fr:
To avoid merge conflicts maybe we should first merge #198 
What is your opinion?

tomschr:
Yes, good idea! :+1:

scls19fr:
If you are ok we can merge it.
And after create a PR for CHANGELOG (both for #198 and for this PR #200).
And an other one about Black

ppkt:
@tomschr Nice PR - I love this formatter :)
I know it's already merged but I have an idea - maybe instead of formatting code base and uploading .diff file, this action should only check whether code is formatted (i.e. exit code from black is 0 - see option `--check`)?

tomschr:
@ppkt Thanks for your idea! :+1: 

Yes, we can do that. :wink: My original idea was to comment on the corresponding pull request to let the contributor know that there is an issue with the code. However, that turned out to be a bit more difficult than I thought.

Probably, the diff is of not real use anyway. It was a start to see how we could use these GH Actions. I'll create an issue for this. Would you review it once I've implemented it? :wink:

PR #202, opened 2019-11-24
Fix #201: Reformat source code with black

### LINKED ISSUE #201: Code formatting using black
Hello,

We now have a Github Action which (should) show diff between current code and code formatted using `black` thanks to PR #200 

See https://github.com/python-semver/python-semver/actions?query=workflow%3A%22Black+Formatting%22

We need now to improve our code base current formatting using black.

Kind regards

tomschr:

* Add `pyproject.toml` containing black configuration
* Add batch image in `README.rst`
* Reformat source code with `black` (version 19.3b0)

Keep in mind, `black` reformats the code quite drastically. That means, all open pull requests will fail to merge. Maybe we should finish and merge pr #138?

---

You will notice, due to #200, the "Check" tab contains now two checks: one for Travis and another one with the GitHub Action for Black. Currently, the black formatting has zero diffs. :+1:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Features
 * :gh:`187` (:pr:`188`): Added logo for python-semver organization and documentation
 * :gh:`191` (:pr:`194`): Created manpage for pysemver
 * :gh:`196` (:pr:`197`): Added distribution specific installation instructions
-
+* :gh:`201` (:pr:`202`): Reformatted source code with black
 
 Bug Fixes
 ---------
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@ Quickstart
 
 A Python module for `semantic versioning`_. Simplifies comparing versions.
 
-|build-status| |python-support| |downloads| |license| |docs|
+|build-status| |python-support| |downloads| |license| |docs| |black|
 
 .. teaser-end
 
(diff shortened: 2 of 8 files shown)


tomschr:
Ahh, the Travis job fails due to "flake8 issues". What should we do with that target? I guess, we should replace it with some "black magic"? :smile:

tomschr:
Actually, it can be solved with this addition in `setup.cfg`:

```ini
max-line-length = 88
exclude =
    .env,
    .eggs,
    .tox,
    .git,
    __pycache__,
    build,
    dist
```

scls19fr:
LGTM... except CHANGELOG 😄

tomschr:
> CHANGELOG is missing

@scls19fr Right. I've added it in commit 1dc8f9c.

tomschr:
Another question: Should we finish and merge pr #138 first before merging this?

scls19fr:
Not sure we reached consensus with #138
So maybe we should first merge this one...
especially because fixing merge issue in #138 shouldn't be difficult.

tomschr:
I'm fine with either one. :wink:  Actually, I was also thinking about the other open pull requests and if we make live of our contributors a bit more difficult. On the other side, some of them are quite old and need to be reviewed (closed, delegated, etc) anyway.

ppkt:
I guess if everyone will run black there should be no conflicts :)

PR #204, opened 2019-11-25
Fix #203: Use --check option for black (infra)

### LINKED ISSUE #203: Improve GH Action for Black to use --check
## Situation

_Originally posted by @ppkt in https://github.com/python-semver/python-semver/pull/200#issuecomment-558041509_:

> I know it's already merged but I have an idea - maybe instead of formatting code base and uploading .diff file, this action should only check whether code is formatted (i.e. exit code from black is 0 - see option `--check`)?

## Proposed Solution

* Remove diff creation
* Add option `--check`

tomschr:

* Remove uploading the diff file as artifact
* Use `--check` to test our source code

---

I haven't provided an entry in the CHANGELOG. It doesn't affect the source code and it's not bound to a specific release, so I thought I skip this. Hope this is ok. :wink:

### DIFF
--- a/.github/workflows/black-formatting.yml
+++ b/.github/workflows/black-formatting.yml
@@ -39,11 +39,5 @@ jobs:
       - name: Run black
         id: black
         run: |
-          black . > project.diff
+          black --check .
           echo "::set-output name=rc::$?"
-
-      - name: Upload diff artifact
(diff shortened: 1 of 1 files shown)


ppkt:
it's too late but it looks good, thanks :)

PR #209, opened 2019-12-08
Fix #208: Introduce VersionInfo.isvalid() function

### LINKED ISSUE #208: Use Case: Introduce a "isvalid()" function to check for valid semver version
## Situation

At the moment, we don't have a function which checks a version string and returns True/False values.

The user is forced to parse the version string and have to check for a `ValueError` exception:

```python
import semver

try:
   version = "..."
   semver.VersionInfo.parse(version)
   # if we are still here, then it's true
except ValueError:
   # seems, version wasn't a good semver version
```

I think, this is a bit cumbersome.

## Proposed Solution

It would be nice if we could do something like this:

```python

version = "..."
if semver.VersionInfo.isvalid(version):
   print("Good version :-)!")
else:
  print("Bad version :-(")
```

---

Open questions:

1. Is this use case useful?
1. Currently, I see `semver.VersionInfo.isvalid()` as a classmethod. Should we better place this function outside the class, like `semver.isvalid()`?
1. Regardless of the location, is the name ok? It resembles the `is...()` functions from the `str` datatype.

tomschr:
This PR fixes #208 and contains:

* `VersionInfo.isvalid(cls, version:str) -> bool`
* Add test case
* Describe function in documentation; also clarified in the "Creating a Version" what happens when the user pass an invalid version string.

As I expect some discussions and as this is still WIP, I haven't completed the CHANGELOG entry yet. :wink: 

---

Update 2019-12-08: Add entry in `CHANGELOG.rst`.

---

Update 2019-12-09: Insert `check` subcommand into `pysemver` command.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@ Features
 * :gh:`191` (:pr:`194`): Created manpage for pysemver
 * :gh:`196` (:pr:`197`): Added distribution specific installation instructions
 * :gh:`201` (:pr:`202`): Reformatted source code with black
+* :gh:`208` (:pr:`209`): Introduce new function :func:`semver.VersionInfo.isvalid`
+  and extend :command:`pysemver` with :command:`check` subcommand
+
 
 Bug Fixes
 ---------
@@ -53,7 +56,6 @@ Features
 * :pr:`165`: Improved code coverage
 * :pr:`166`: Reworked :file:`.gitignore` file
 * :gh:`167` (:pr:`168`): Introduced global constant :data:`SEMVER_SPEC_VERSION`
-
 
 Bug Fixes
 ---------
--- a/docs/pysemver.rst
+++ b/docs/pysemver.rst
@@ -86,6 +86,29 @@ you get an error message and a return code != 0::
    ERROR 1.5 is not valid SemVer string
 
 
+pysemver check
+~~~~~~~~~~~~~~
+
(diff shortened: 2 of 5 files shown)


scls19fr:
I'm fine with both with naming and the fact it's a class method.

tomschr:
Ok, I've updated the commit and inserted an entry for the CHANGELOG.

scls19fr:
Waiting some days to have others @python-semver/reviewers feedback before merging.

tomschr:
@ppkt Good point Karol! :+1: I've never thought about this. :smile: 

Ok, will amend this PR

tomschr:
@ppkt Karl, I've added this subcommand now and extended the manpage. However, I've realized that I added is as `is-valid` instead of `check`. I will fix that.

tomschr:
I've also added testcase to test the `check` subcommand. Feel free to comment. :wink:

scls19fr:
great idea to also add such command to cli also!

tomschr:
@scls19fr, @ppkt I realized, the implementation of the subcommands are a bit suboptimal. I will optimize this in a separate commit. :wink: 

Any other wishes, comments, ideas, etc.?

ppkt:
Looks really great, thank you :)

scls19fr:
Thanks @tomschr for this new feature. Thanks @ppkt for your review.

PR #211, opened 2019-12-09
Fix #210: Add classmethod VersionInfo.coerce

### LINKED ISSUE #210: Use case: Introduce VersionInfo.coerce
## Situation

Currently, if you parse a version string into a valid semver version, the version string has to consist of at least the major, minor, and patch parts.

However, in some use cases some parts could be missing, for example the patch part.

## Proposed Solution

I propose a classmethod `VersionInfo.coerce` with the following signature:

```python
@classmethod
def coerce(cls, version):
    """
    Convert an incomplete version string into a semver-compatible VersionInfo
    object

    * Tries to detect a "basic" version string (``major.minor.patch``).
    * If not enough components can be found, missing components are
      set to zero to obtain a valid semver version.

    :param str version: the version string to convert
    :return: a tuple with a :class:`VersionInfo` instance (or ``None``
        if it's not a version) and the rest of the string which doesn't
        belong to a basic version.
    :rtype: tuple(:class:`VersionInfo` | None, str)
    """
```

I've chosen `coerce` as it is a common function name in other semver implementations, for example, [node-semver](https://github.com/npm/node-semver/blob/master/semver.js#L1548).

An example session:

```python
>>> import semver
>>> semver.VersionInfo.coerce("v1.2")
(VersionInfo(major=1, minor=2, patch=0, prerelease=None, build=None), '')
>>> semver.VersionInfo.coerce("3.5rc1")
(VersionInfo(major=3, minor=5, patch=0, prerelease=None, build=None), 'rc1')
```

Questions:

* Do you you think such a function could be useful?
* Should the `coerce` function accept an integer value (=major)?  So `coerce(2)` would be the same as `coerce("2")` which would return `VersionInfo(major=2, minor=0, patch=0, prerelease=None, build=None)`.
* Anything else you miss, would change, ...?

## See also
This issue is related to #137

tomschr:
This PR fixes #210 and contains the following changes:

* Add `VersionInfo.coerce(cls, version:str) -> tuple`

Convert an incomplete version string into a semver-compatible `VersionInfo` object:

* Tries to detect a "basic" version string (``major.minor.patch``).
* If not enough components can be found, missing components are set to zero to obtain a valid semver version.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Features
 * :gh:`201` (:pr:`202`): Reformatted source code with black
 * :gh:`208` (:pr:`209`): Introduce new function :func:`semver.VersionInfo.isvalid`
   and extend :command:`pysemver` with :command:`check` subcommand
+* :gh:`210` (:pr:`211`): Added classmethod :func:`semver.VersionInfo.coerce`
 
 
 Bug Fixes
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -213,6 +213,35 @@ Depending which function you call, you get different types
     {'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None}
 
 
+Parsing Incomplete Versions
+---------------------------
+
+If a version is incomplete (some parts are missing) this would lead to an
+exception during the parsing process. As such, you cannot get a valid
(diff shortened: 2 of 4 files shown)


tomschr:
Closing in favor of #215

PR #214, opened 2019-12-10
Improve pysemver subcommands

tomschr:
This PR is not based on a specific issue, but is a refactoring effort to improve maintenance of the `pysemver` subcommands:

* Use separate `cmd_*` functions for each subcommand (basically, I've moved the if...elif clauses from `process` into separate functions).
* Replace `which` with `func` keyword.
* Use `func` keyword to store the specific `cmd_*` function.
* Adapt tests.
* (Re)format with `black`.
* Clarify how to interpret return code in manpage (add new section "Return Code").

=> Easier to extend subcommands

### DIFF
--- a/docs/pysemver.rst
+++ b/docs/pysemver.rst
@@ -133,16 +133,38 @@ to indicates which is the bigger version:
 * ``0`` if both versions are the same,
 * ``1`` if the first version is greater than the second version.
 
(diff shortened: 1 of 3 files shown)


tomschr:
@python-semver/reviewers any comments? :wink:

PR #229, opened 2020-03-15
Deprecate module level functions

### LINKED ISSUE #225: Decide to deprecate module level functions?
## Situation

Currently, we have mostly functions which are either on the module level (`semver.next_version`, `semver.bump_*`) and almost the same on the `VersionInfo` level (`semver.VersionInfo.next_version`, `semver.VersionInfo.bump_*`).

I know, they both serve distinctive purposes: the first expects and returns a string whereas the second needs a `VersionInfo` object.

However, from an implementation perspective, I'm wondering if this is double work. It looks inefficient to me when the `VersionInfo` methods need to convert the object first into a string, call the modul level function, and convert it back to a `VersionInfo` object.

IMHO, we should improve this situation.

## Suggested Solution

The question boils down to this: should we treat `VersionInfo` as **first class object**?

I have the impression this isn't currently the case. The developer can decide between two implementation which does mostly the same. The only difference is, one implementation use strings, the other use a "detour" with an `VersionInfo` object.

If we would treat `VersionInfo` as our one and only solution, that would mean we have to move the implementation into the methods. The module level functions would serve as "helper functions".

Or... maybe we could think about a more radical approach? Are the module level functions really necessary? Should we deprecate them? As we can convert a `VersionInfo` object into a string through `str(versionobject)`, it seems to me, they are obsolete...

I see the following benefits of this approach:

* Focus on the `VersionInfo` class, all the magic would happen there. :wink: 
* Avoids writing two functions which mostly implements the same functionality, but with different types.
* Simplifies testing.
* Focus more on objects than strings (which can't be controlled).
* Improves the project to be more slim and well-arranged.
* Makes conversion between objects obsolete. Without conversion, it would have a improved effect on performance (maybe only slightly, but...).

If we want to follow this path, we would need:

* Use the `warning` module to issue a deprecation warning in affected module functions.
* Document the change for our upcoming major release. :wink:
* Release 2.9.2 of semver to implement the warnings
* Make the whole removal in release 3 only.

@gsakkis, @scls19fr  Does that make sense to you? Do I miss an important point? I would really like to hear your view on this. :heart:

tomschr:
This PR fixes #225. 

For example, the `VersionInfo.bump_major()` method calls `semver.parse_version_info` and `semver.bump_major` instead of having its own implementation. This has several drawbacks:

* the VersionInfo class depends on other, "remote" function outside the class.
* the class is not "self-contained".
* it makes deprecating module level functions (and removing them) harder because of these dependencies.

As such, I had to move some implementation into the `VersionInfo` class. With this approach, we can deprecate the module level function without triggering a false positive when calling a method in `VersionInfo`.

* Add test cases
  - Test also missing `finalize_version`
  - Test the warning more thouroughly with `pytest.warns` instead of just `pytest.deprecated_call`
* In `setup.cfg`, add deprecation warnings filter for pytest
* Implement DeprecationWarning with warnings module and the new decorator `deprecated`
* Output a DeprecationWarning for the following functions:
    - `semver.bump_{major,minor,patch,prerelease,build}`
    - `semver.finalize_version`
    - `semver.format_version`
    - `semver.parse`
    - `semver.parse_version_info`
    - `semver.replace`
* Move module level implementation into `VersionInfo`
   *  Move implementations:
    
      - `semver.format_version` -> ~`semver.VersionInfo.format_version`~ `semver.VersionInfo.__str__`
      - Move `_REGEX` into `VersionInfo` class
      - Move `_LAST_NUMBER` into `VersionInfo` class
      - Move also `_increment_string` and implement it as a staticmethod
      - `semver.parse` -> `semver.VersionInfo.parse`
      - `semver.bump_*` -> `semver.VersionInfo.bump_*`
    
  *  Change implementation by calling VersionInfo methods (and not the other way around):

     - `semver.compare`
     - `semver.finalize_version`
     - `semver.replace`
     - `semver.parse_version_info`

* Introduce new public functions:
     - `semver.VersionInfo.to_dict` (from former `_asdict`)
     - `semver.VersionInfo.to_tuple` (from former `_astuple`)
     - Keep `_asdict` and `_astuple` as a (deprecated) function for  compatibility reasons

* Update CHANGELOG.rst

* Update usage documentation:
      - Move some information to make them more useful for the reader
      - Add deprecation warning
      - Explain how to replace deprecated functions
      - Explain how to display deprecation warnings from semver

* Improve documentation of deprecated functions
     - List deprecated module level functions
     - Make recommendation and show equivalent code
     - Mention that deprecated functions will be replaced in  semver 3. That means, all deprecated function will be still available in semver 2.x.y.

@gsakkis As you were also involved, I'd like to hear your opinion too. :wink:

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -3,9 +3,11 @@ from __future__ import print_function
 
 import argparse
 import collections
-from functools import wraps
+from functools import wraps, partial
+import inspect
 import re
 import sys
+import warnings
 
 
 __version__ = "2.9.1"
@@ -14,28 +16,6 @@ __author_email__ = "k-bx@k-bx.com"
 __maintainer__ = ["Sebastien Celles", "Tom Schraitle"]
 __maintainer_email__ = "s.celles@gmail.com"
 
-_REGEX = re.compile(
-    r"""
-        ^
-        (?P<major>0|[1-9]\d*)
-        \.
-        (?P<minor>0|[1-9]\d*)
-        \.
-        (?P<patch>0|[1-9]\d*)
-        (?:-(?P<prerelease>
-            (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
-            (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
-        ))?
-        (?:\+(?P<build>
-            [0-9a-zA-Z-]+
-            (?:\.[0-9a-zA-Z-]+)*
-        ))?
-        $
-        """,
-    re.VERBOSE,
-)
-
-_LAST_NUMBER = re.compile(r"(?:[^\d]*(\d+)[^\d]*)+")
 
 #: Contains the implemented semver.org version of the spec
 SEMVER_SPEC_VERSION = "2.0.0"
@@ -47,9 +27,65 @@ if not hasattr(__builtins__, "cmp"):
         return (a > b) - (a < b)
 
 
+def deprecated(func=None, replace=None, version=None, category=DeprecationWarning):
+    """
+    Decorates a function to output a deprecation warning.
+
+    This function will be removed once major version 3 of semver is
+    released.
+
+    :param str replace: the function to replace (use the full qualified
+        name like ``semver.VersionInfo.bump_major``.
+    :param str version: the first version when this function was deprecated.
+    :param category: allow you to specify the deprecation warning class
+        of your choice. By default, it's  :class:`DeprecationWarning`, but
+        you can choose :class:`PendingDeprecationWarning``or a custom class.
+    """
+
+    if func is None:
+        return partial(deprecated, replace=replace, version=version, category=category)
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        msg = ["Function '{m}.{f}' is deprecated."]
+
+        if version:
+            msg.append("Deprecated since version {v}. ")
+        msg.append("This function will be removed in semver 3.")
+        if replace:
+            msg.append("Use {r!r} instead.")
+        else:
+            msg.append("Use the respective 'semver.VersionInfo.{r}' instead.")
+
+        # hasattr is needed for Python2 compatibility:
+        f = func.__qualname__ if hasattr(func, "__qualname__") else func.__name__
+        r = replace or f
+
+        frame = inspect.currentframe().f_back
+
+        msg = " ".join(msg)
+        warnings.warn_explicit(
+            msg.format(m=func.__module__, f=f, r=r, v=version),
+            category=category,
+            filename=inspect.getfile(frame.f_code),
+            lineno=frame.f_lineno,
+        )
+        # As recommended in the Python documentation
+        # https://docs.python.org/3/library/inspect.html#the-interpreter-stack
+        # better remove the interpreter stack:
+        del frame
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+@deprecated(version="2.9.2")
 def parse(version):
     """
     Parse version to major, minor, patch, pre-release, build parts.
+
+    .. deprecated:: 2.9.2
+       Use :func:`semver.VersionInfo.parse` instead.
 
     :param version: version string
     :return: dictionary with the keys 'build', 'major', 'minor', 'patch',
@@ -69,17 +105,7 @@ def parse(version):
     >>> ver['build']
     'build.4'
     """
-    match = _REGEX.match(version)
-    if match is None:
-        raise ValueError("%s is not valid SemVer string" % version)
-
-    version_parts = match.groupdict()
-
-    version_parts["major"] = int(version_parts["major"])
-    version_parts["minor"] = int(version_parts["minor"])
-    version_parts["patch"] = int(version_parts["patch"])
-
-    return version_parts
+    return VersionInfo.parse(version).to_dict()
 
 
 def comparator(operator):
@@ -110,6 +136,29 @@ class VersionInfo(object):
     """
 
     __slots__ = ("_major", "_minor", "_patch", "_prerelease", "_build")
+    #: Regex for number in a prerelease
+    _LAST_NUMBER = re.compile(r"(?:[^\d]*(\d+)[^\d]*)+")
+    #: Regex for a semver version
+    _REGEX = re.compile(
+        r"""
+            ^
+            (?P<major>0|[1-9]\d*)
+            \.
+            (?P<minor>0|[1-9]\d*)
+            \.
+            (?P<patch>0|[1-9]\d*)
+            (?:-(?P<prerelease>
+                (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
+                (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
+            ))?
+            (?:\+(?P<build>
+                [0-9a-zA-Z-]+
+                (?:\.[0-9a-zA-Z-]+)*
+            ))?
+            $
+        """,
+        re.VERBOSE,
+    )
 
     def __init__(self, major, minor=0, patch=0, prerelease=None, build=None):
         self._major = int(major)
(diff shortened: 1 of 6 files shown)


semver.py
```diff
@@ -47,10 +28,44 @@ def cmp(a, b):
...
+def deprecated(replace=None):
+    """
+    Decorates a function to output a deprecation warning.
+
+    This function will be removed once major version 3 of semver is
+    released.
+    """
+    # we can't use the 'nonlocal' keyword in Python2, so we need
+    # to circumvent that with a dictionary:
+    r = {"r": replace}
```
gsakkis: `nonlocal` is needed only when you reassign a variable (`replace = ...`), not when you just access it. Thus lines 40 and 46 can leave and replace `r=replace` in line 52 with `r=replace or func.__name__`

semver.py
```diff
@@ -344,6 +386,31 @@ def isvalid(cls, version):
         except ValueError:
             return False
 
+    @staticmethod
+    def format_version(major, minor, patch, prerelease=None, build=None):
```
gsakkis: Does this really need to be a public method? I don't find it particularly useful; I'd just inline it in `__str__` and suggest `str(VersionInfo(...))` as substitute of the deprecated `format_version` function.

semver.py
```diff
@@ -47,10 +28,44 @@ def cmp(a, b):
...
+    return decorator
+
+
+@deprecated()
 def parse(version):
     """
     Parse version to major, minor, patch, pre-release, build parts.
 
+    .. deprecated:: 2.9.2
+       Use :func:`semver.VersionInfo.format_parse` instead.
```
gsakkis: There's no `format_parse`.

semver.py
```diff
@@ -658,30 +729,48 @@ def bump_build(version, token="build"):
...
+    return str(VersionInfo.parse(version).bump_build(token))
 
 
+@deprecated()
 def finalize_version(version):
     """
     Remove any prerelease and build metadata from the version.
 
+    .. deprecated:: 2.9.2
+       Use :func:`semver.VersionInfo.bump_format_version` instead.
```
gsakkis: There's no `bump_format_version`

semver.py
tomschr: Good point! :+1: I removed the above function and added the `str(VersionInfo(...))` fragment into the original module level function `format_version`.

semver.py
tomschr: Right, that was a copy-and-paste error. Fixed.

semver.py
tomschr: True, I was mislead by this nonlocal keyword. Replaced it with your suggestion. :+1:

semver.py
tomschr: Right, it was a bad typo. I'm wondering what we should do with `semver.finalize_version`. Probably, there should be a `semver.VersionInfo.finalize_version` method, right?
Or do you see another option?

tomschr: After thinking about this, I guess a `semver.VersionInfo.finalize_version` would be helpful. I think, the signature would be:

```python
def finalize_version(self):
    ...
```

However, I'm not sure about what to return:

1. Set prerelease and build to None and return `self`.
2. Create a new instance with major, minor, and patch. So efficiently, it does not touch the current version.

What do you think should be the preferred way? (IMHO, I tend to option 1)

scls19fr:
conflict with semver.py need to be fixed.

tomschr:
@scls19fr I fixed the conflicts now.

Please wait as there is one thing to do: I would like to improve the documentation. I think, this still needs a note or some further explanations.

I'll keep you informed.

scls19fr:
I have no idea about what is the best... deprecating module function in major 3.x.x or in 4.x.x.

Anyway documenting is the best we can do currently.

tomschr:
@scls19fr 
> I have no idea about what is the best... deprecating module function in major 3.x.x or in 4.x.x.

No problem. In that case, we could deprecate them in 4.x.y to be on the safe side. That may give projects a little bit more time and we could address @ppkt concerns.

---

There is another issue which just came into my mind: should we make the `_asdict()` method "public" (=removing the underscore from the function)?

The reason for this is it could help projects which still uses `semver.parse()` and using dictionaries. That could ease the migration path.

scls19fr:
I'm wondering if `todict` won't be a better name.

tomschr:
Yep, that's probably a better name. :+1: I will do the change and add also some tests.

scls19fr:
or `to_dict` as in Pandas

scls19fr:
see in https://pandas.pydata.org/pandas-docs/stable/reference/frame.html
`to_...` methods

tomschr:
Ok, even better. I'll use `to_dict`

tomschr:
You can find my current implementation of `semver.VersionInfo.to_dict`  in commit 9c89a53

scls19fr:
I wonder if we shouldn't even have
- `to_tuple`
- `to_string`
methods

so it will be easier to know using tab completion how to transform a `VersionInfo`
simply by writing `VersionInfo.to_<tab>`

tomschr:
Actually, I thought of that too. :smile: 

I think, the `to_tuple` could be useful. :+1: 

However, I'm not sure about the `to_string`. I think, in this case we have already a solution: `str(version)`. IMHO we shouldn't offer a second way ("There should be one-- and preferably only one --obvious way to do it.").

scls19fr:
ok! let's forget `to_string`

tomschr:
@scls19fr I've updated the documentation, see for details commit c4ac0ec.

All changes are in section "Using semver". As the changes are a bit spread, I would recommend to build the documentation with `tox -e docs` and going through the HTML to get an impression.

If you are missing some important information, let me know. After your approval, I would squash the commits to have a clean history.

semver.py
tomschr: After a second thought, I've implemented option 2. That would be consistent with, for example, the `bump_*` methods. I've marked this discussion as resolved.

tomschr:
@scls19fr Thanks for the approval. :+1:  However, before I can squash and merge it, I found some other issues which I would like to ask.

In commit 1d8c415, I've moved the `_LAST_NUMBER` variable and the function `_increment_string` into the `VersionInfo` class. IMHO, now the class contains everything that is needed and it is not dependent from another "external" function. Would that be fine for you?

Another question is regarding the warnings. As I've improved the `deprecated` decorator in commit 08c71eb I'm not sure anymore if I've introduced some issues.

For example, I've tried to display the warnings, however this seems not work:

```shell
$ python3  -c "import semver; print(semver.parse_version_info('3.4.5'))"
3.4.5
```

However, when I add the `-W` option with `d` (=default), then I can see them:

```shell
$ python3 -Wd -c "import semver; print(semver.parse_version_info('3.4.5'))"
<string>:1: DeprecationWarning: Function 'semver.parse_version_info' is deprecated. Deprecated since version 2.9.2. Use 'semver.VersionInfo.parse' instead.
3.4.5
```

@ppkt, @scls19fr do you know if this is an expected behaviour? Is the correct? Not sure... Do you know this?

Thanks for your help!

semver.py
```diff
@@ -47,10 +27,69 @@ def cmp(a, b):
...
+        you can choose :class:`PendingDeprecationWarning``or a custom class.
+    """
+
+    def decorator():
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            msg = ["Function 'semver.{f}' is deprecated."]
+
+            if version:
+                msg.append("Deprecated since version {v}.")
```
ppkt, line 51: Maybe you could as well add information when deprecated function will be removed? Just a suggestion :)

semver.py
```diff
@@ -47,10 +27,69 @@ def cmp(a, b):
...
+
+            if replace:
+                msg.append("Use {r!r} instead.")
+            else:
+                msg.append("Use the respective 'semver.VersionInfo.{r}' instead.")
+
+            # warnings.simplefilter('always', category)
+            frame = inspect.currentframe().f_back
```
ppkt, line 59: May I ask what happens here?

semver.py
tomschr: Good Idea! :+1: Will add it. :-)

semver.py
tomschr: Of course! This is needed to get the frame object which contains the line number, the filename etc. from the originated deprecated function. This is then passed to the `warnings.warn_explict` around line 63.

ppkt:
@tomschr yes, it seems that by default DeprecationWarnings are [ignored](https://docs.python.org/3/library/warnings.html#warning-categories): 
> Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in `__main__`).

I'm not sure about meaning of last sentence however

tomschr:
@ppkt Thanks for your analysis. Much appreciated. :+1: 

> yes, it seems that by default DeprecationWarnings are [ignored](https://docs.python.org/3/library/warnings.html#warning-categories):

Yes, that was also my findings, but I wasn't sure.
 
> > Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in `__main__`).
>
> I'm not sure about meaning of last sentence however

Hmn, maybe they mean it something like this:

```python
import semver  # not triggered in the library?

if __name__ == "__main__":
   # Warning is triggered here
```

Just a theory. I think, I should mention the `-W`/`-Wd` option in the documentation to make developers aware of this.

tomschr:
Yes, my suspicion seems to be correct. Here is a short test code:

```python
#!/usr/bin/python3

import warnings
import semver

def fourtytwo():
    warnings.warn("Oh no, we found the answer!", category=UserWarning)
    return 42

if __name__ == "__main__":
  print(semver.bump_major('3.4.5'))
  print(fourtytwo())
```

When I run this script, I get:

```shell
$ ./fortytwo.py
4.0.0
./fortytwo.py:7: UserWarning: Oh no, we found the answer!
  warnings.warn("Oh no, we found the answer!", category=UserWarning)
42
```

which is exactly the result predicted from the Python documentation.

When I change the shebang line and add the `-Wd` option, I get all:

```shell
$ ./fortytwo.py
./fortytwo.py:11: DeprecationWarning: Function 'semver.bump_major' is deprecated. Deprecated since version 2.9.2.  This function will be removed in semver 3. Use the respective 'semver.VersionInfo.None' instead.
  print(semver.bump_major('3.4.5'))
4.0.0
./fortytwo.py:7: UserWarning: Oh no, we found the answer!
  warnings.warn("Oh no, we found the answer!", category=UserWarning)
42
```

tomschr:
@scls19fr @ppkt @gsakkis  I've squashed the commits, but added all necessary information into the commit message.

I'm finished. The documentation is updated, the test cases works and even the coverage reports 100%! :tada: 
(However, for some odd reason, Travis seems to cease its work. :disappointed: Any hints?)

Could you have a final look if everything is ok? I will merge it soon.

semver.py
ppkt: Ah, I see, thanks :)

tomschr:
As the changes are a bit big, I will leave you some more time. If nobody objects, I will merge it on Tuesday. :wink:

Issue #213, opened 2019-12-10
Consider type hints for 3.x.y release?


tomschr:
## Situation

Python 3.5 introduced the [`typing`](https://docs.python.org/3/library/typing.html#module-typing) module. As soon as we are sunsetting the 2.9.x release of semver, it raises the question if we should use type hints for our 3.x.y release of semver.

Currently, we don't use it for our functions and methods.

To avoid any misinterpretations, the `typing` module contains this note:

> The Python runtime does not enforce function and variable type annotations. They can be used by third party tools such as type checkers, IDEs, linters, etc.

To start discussion and to get a better picture where we stand on this topic, I've opened this issue. :wink: 

## Example

This is a short example of a type hint for the `parse` function:

```python
def parse(version: str) -> dict:
    """
    Parse version to major, minor, patch, pre-release, build parts.

    :param version: version string
    :return: dictionary with the keys 'build', 'major', 'minor', 'patch',
     and 'prerelease'. The prerelease or build keys can be None
     if not provided
```

Bernát Gábor recommends that “type hints should be used whenever unit tests are worth writing.”

## Pros and Cons

I've collected some pros and cons, probably there are more that I haven't listed here. This list is by no means exhaustive. Feel free to amend it.

### Pros

* Helps to catch type errors when testing with mypy.
* Could support IDEs to provide a better popup help.
* Improves documentation when using the [`sphinx-autodoc-typehints`](https://github.com/agronholm/sphinx-autodoc-typehints) plugin.
*  Help you as a developer to write better code.

### Cons

* A slight impact of performance.
* Needs another dependency when checking for type hints (`mypy`).
* Needs a bit more time to add it.
* Works best with modern Python (version >= 3.5).

## Scope

If we want this, it will only be implemented in the 3.x.y release of semver. Older releases (2.9.x) are not covered by this issue and won't be modified.

Furthermore, we need to create a `semver.pyi` file with the command `stubgen` (integrated in mypy).

## See also

* https://realpython.com/lessons/pros-and-cons-type-hints/
* https://www.python.org/dev/peps/pep-0483/
* https://www.bernat.tech/the-state-of-type-hints-in-python/
* https://www.python.org/dev/peps/pep-0561/ (Distributing and Packaging Type Information)

ppkt:
I'm a huge fan of typing in Python. I wasn't aware about impact on performance but I think every library should use them. As far as I know, Python 3.4 is no longer maintained so we can drop support for it.
[...]

PR #230, opened 2020-03-16
Add version information in some functions

tomschr:
This PR contains some minor fixes in docstrings only:

* Use `.. versionadded::` RST directive in docstrings to make it more visible when something was added
* Minor wording fix in docstrings (versions -> version strings)

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -307,8 +307,12 @@ prerelease='pre.2', build='build.4')
         return parse_version_info(version)
 
     def replace(self, **parts):
-        """Replace one or more parts of a version and return a new
-          :class:`semver.VersionInfo` object, but leave self untouched
+        """
+        Replace one or more parts of a version and return a new
+        :class:`semver.VersionInfo` object, but leave self untouched
+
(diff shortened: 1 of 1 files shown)


semver.py
```diff
@@ -433,7 +442,7 @@ def _compare_by_keys(d1, d2):
 
 def compare(ver1, ver2):
     """
-    Compare two versions.
+    Compare two versions strings.
 
     :param ver1: version string 1
```
ppkt: A quick question / suggestion - maybe in such case we should also add typing? Like for instance: 
```suggestion
    :param str ver1: version string 1
```
IDEs (like PyCharm) would love that feature :)

tomschr: Generally you are right. :+1:
However, I'm not sure. We have already issue #213 which is targeted for major version 3. So maybe, I'd suggest to wait until we can release the next major release? Would that be ok?

ppkt: Yeah, sure :)

tomschr:
As this PR is only about docstrings, it shouldn't create any problems, so merging it.

Issue #160, opened 2019-10-07
Deprecate min_ver/max_ver and use builtin max()/min() functions


tomschr:
## Situation

~The `min_ver` and `max_ver` functions have currently this signature:~

```python
def max_ver(ver1, ver2):
    # ...
def min_ver(ver1, ver2):
    # ...
```

~In other words, these functions allow to **compare exactly two versions**. However, there may be situations, where you have a *list of versions*. This use case could not be done with the current implementation and needs a cumbersome for-loop.~

As we have comparison operators, there is no need for this functions anymore. Of course, this works only for VersionInfo types, not strings. If someone still wants to compare strings, we could document a replacement.

Action Items:

* [ ] add deprecation warning for both functions
* [ ] document replacement; maybe something like: `max(map(semver.VersionInfo.parse, ['1.1.0', '1.2.0', '2.1.0', '0.5.10', '0.4.99']))`

-----

Old issue:

## Suggested Solution

To allow to compare more than two versions, it would be useful to change the above signature like this:

```python
def max_ver(ver1, ver2, *others):
    # ...
def min_ver(ver1, ver2, *others):
    # ...
```

This would resemble the builtin functions `min` and `max`. The above functions could be used like this:

```python
from semver import max_ver

m = max_ver("3.0.0", "2.0.0", "1.0.0", "4.0.0")
print(m)
# => "1.0.0"
```

Benefits:

* No incompatible changes to API; it's still possible to use two arguments
* Only minimal changes in the implementation

## Alternative Solution

The proposed solution may has some drawbacks:

* it doesn't allow to pass an iterable (like the builtin functions `min` and `max`).
* could hit a performance bottleneck

If you don't like the proposed solution, we could:

* leave the current implementation of `min_ver` and `max_ver` untouched, or,
* allow iterables in  `min_ver` and `max_ver` (would introduce incompatible API changes),
* introduce additional functions with the suffix `_iter` which allows to optimize for iterables. The signature could be:

```python
def max_ver_iter(iterable, *, default=None, key=None):
    """Returns the biggest version from an iterable

   :param iterable: an iterable with version strings
   :param default: specifies an object to return if the provided iterable is empty.
   :param key: specifies a one-argument ordering function like that used for list.sort()
   :raises: ValueError (If the iterable is empty and default is not provided)
    """
```

## Questions

* Does it make sense?
* Should we raise something if the `*others` tuple contains an invalid version?
[...]

PR #239, opened 2020-04-21
Fix #236: add missing deprecated functions

### LINKED ISSUE #236: Deprecate missing module-level functions (like semver.compare, semver.match, ...)
# Situation

We deprecated most of our module-level functions, but it seems, I forgot `semver.match`.

# Proposal

I assume, this function should also be added to the list of deprecated function. It also should be moved to the `semver.VersionInfo` class.

@python-semver/reviewers Any objections?

---

There are some other module-level functions:

* `semver.compare`
* `semver.max_ver`
* `semver.min_ver`

@gsakkis: What should we do with these? Deprecate them too and move them to the `semver.VersionInfo` class?

tomschr:
This PR fixes #236 and contains the following changes:

Deprecate:

* [x] `semver.compare` -> `semver.VersionInfo.compare`
* [x] `semver.match` -> `semver.VersionInfo.match`

Change:

* Implementation of `semver.max_ver` and `semver.min_ver`

----

~The implementation in this PR is a minimal implementation. I could have done more, but~ I would like to hear your opinion. I have two questions:

* Should we really move the functions `semver.max_ver` and `semver.min_ver` into the class? Especially if we maybe want to change them later, see #160.

* What about the "private" functions `_nat_cmp` ~and `_compare_by_keys`~? Should we move them into the class as well.

@python-semver/reviewers What do you think?


tomschr:
I've found out a method to get rid of `__compare_by_keys` alltogether. The algorithm is hidden in the `VersionInfo.compare` method. That would influence the comparision operators too.

For example, rewriting the `__eq__` method, it ends up like this:

```python
def __eq__(self, other):
    return self.compare(other) == 0
```

IMHO this is much nicer than the original one:

```python
def __eq__(self, other):
    return _compare_by_keys(self.to_dict(), _to_dict(other)) == 0
```

The `VersionInfo.compare` method take into account if `other` is another `VersionInfo` instance, a dict, a tuple/list, or a string.

I'll push another commit to demonstrate that.

tomschr:
If nobody objects, I would merge it tomorrow. 😉

PR #242, opened 2020-04-26
Add __all__ to provide explict index of semver

tomschr:
This PR is not based on a specific issue.

This PR just defines default behavior when executing `from semver import *` (although it's considered bad coding practice).

See also:
https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

---

As this is kind of minor change, I would merge it on Monday if nobody objects.

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -10,12 +10,44 @@ import sys
 import warnings
 
 
-__version__ = "2.10.0"
+__version__ = "2.10.1"
 __author__ = "Kostiantyn Rybnikov"
 __author_email__ = "k-bx@k-bx.com"
 __maintainer__ = ["Sebastien Celles", "Tom Schraitle"]
 __maintainer_email__ = "s.celles@gmail.com"
 
+#: Our public interface
+__all__ = (
+    #
+    # Module level function:
+    "bump_build",
+    "bump_major",
+    "bump_minor",
+    "bump_patch",
+    "bump_prerelease",
+    "compare",
+    "deprecated",
+    "finalize_version",
+    "format_version",
+    "match",
+    "max_ver",
+    "min_ver",
+    "parse",
+    "parse_version_info",
+    "replace",
+    #
+    # CLI interface
+    "cmd_bump",
+    "cmd_check",
+    "cmd_compare",
+    "createparser",
+    "main",
+    "process",
+    #
+    # Constants and classes
+    "SEMVER_SPEC_VERSION",
+    "VersionInfo",
+)
 
 #: Contains the implemented semver.org version of the spec
 SEMVER_SPEC_VERSION = "2.0.0"


scls19fr:
Sorry but I don't understand the purpose of this PR.

What happens if __all__ is not defined?
What happens if our definition of __all__ contains more functions, classes than default?
On this opposite what happens if our definition of __all__ contains less functions, classes than default?

tomschr:
@scls19fr No problem, I had better explained myself.

> What happens if `__all__` is not defined?

Not much. :wink: It will only have an effect, if you import semver with `from semver import *`. Then, and only then, it will be considered. If this variable is **not defined** at all, it will import *everything* from the semver module, except private objects (starting the name with an underscore).

See scenario 1 below.

> What happens if our definition of `__all__` contains more functions, classes than default?

What do you mean "than default"? I guess, you mean when the variable contains names which are not defined inside semver?` Then it will give you an AttributeError (`module 'semver' has no attribute 'foo'`)

See scenario 3.

> Oppositely, what happens if our definition of `__all__` contains less functions, classes than default?

That is possible and is used/recommended for other modules/packages as well. I did that with the above list. Usually you want to export only objects from semver itself, not third-party modules (like `sys`, `re` etc.).

See Scenario 2.

You can test all three scenarios with the following basic module `foo.py`:

```python
import sys
import os

if False:  # (1)
    __all__ = ("bar",
           # "zoo",  # (2)
           # "notfound",  # (3)
           )

def bar():  ...

def zoo():  ...

def _private():  ...
```

Execute it with:

    python3 -c 'from foo import *; print(locals())'

1. Scenario: No `__all__` defined

    Comment \#1 stays as `False` I get (I only list non-private names):

        {..., 'sys': <module 'sys' (built-in)>, 'os': <module 'os' from '/usr/lib64/python3.6/os.py'>, 'bar': <function bar at 0x7f46e857b620>, 'zoo': <function zoo at 0x7f46e857b6a8>}

   As you can see, it exports everything except the `_private` function.

2. Scenario: With `__all__` defined

   Change the `if` clause in comment \# 1 to `True`, execute it and you will get:

       {...  'bar': <function bar at 0x7fab24fe2620>}

  As you can see, `__all__` exports only `bar`, but not `zoo` (because it's not listed inside `__all__`, it's comment out in \# 2). Furthermore, the result doesn't contain `sys` and `os` (as opposed in scenario 1). However, the `__all__` variable doesn't prevent someone from importing "hidden" object directly with, for example, `from foo import zoo`. That would still be possible.

3. Scenario: With `__all__` defined, but with an unknown object

   Change the `if` clause in comment \# 1 to `True`, remove comment from \# 3 and execute it. This time you get an exception:

        AttributeError: module 'foo' has no attribute 'notfound'
 
--- 

You're right to wonder what this is all about. According to the [StackOverflow](http://stackoverflow.com/a/2187636/434799) answer, it has two purposes:


> 1. Anybody who reads the source will know what the exposed public API is. It doesn't prevent them from poking around in private declarations, but does provide a good warning not to.
>
> 2.  When using from mod import *, only names listed in __all__ will be imported. This is not as important, in my opinion, because importing everything is a really bad idea.

It may be of less importance, but it kind of slightly protects from polluting the global namespace. For example, I've read somewhere that Sphinx will read this variable and take it into account when generating the API documentation.

tomschr:
In the book "Effective Python" from Brett Slatkin, he writes in page 181 about "Stable APIs" (emphasized by me):

> Python can limit the surface area exposed to API consumers by using the `__all__` special
> attribute of a module or package. The value of `__all__` is a list of every name to export from
> the module **as part of its public API**.

PR #243, opened 2020-04-26
Implement __getitem__ for #138

tomschr:
This PR is a fresh implementation of #138 and contains:

* Add `__getitem__` to `VersionInfo` class
* Add test cases
* Add user documentation
* Extend CHANGELOG

----

@tlaferriere, @bittner: As pr #138 was a bit older and a lot of things have changed now, I've created a fresh new branch so we won't get any conflicts. On top of that, I've added the implementation. If you would like to have a look, that would be great. :+1:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,8 @@ Version 2.10.0 (WIP)
 Features
 --------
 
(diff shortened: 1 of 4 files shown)


tomschr:
As nobody seem to object, I'd merge it.

PR #245, opened 2020-04-26
Fix #244: Allow list & str for comparison

### LINKED ISSUE #244: Cannot compare with list or string
# Situation

A `VersionInfo` instance has the ability to compare with another `VersionInfo` instance, a tuple, or a dict :

```python
>>> import semver
>>> v1 = semver.VersionInfo.parse("1.2.3")
>>> v1 < (1, 2, 4)
True
```

However, it is not possible for a list or a string:

```python
>>> v1 < [1, 2, 4]
...
TypeError: other type <class 'list'> must be in (<class 'semver.VersionInfo'>, <class 'dict'>, <class 'tuple'>)
>>> v1 < "1.2.4"
...
TypeError: other type <class 'str'> must be in (<class 'semver.VersionInfo'>, <class 'dict'>, <class 'tuple'>)
```

# Situation

After analyzing the situation above, I found the `comparator` which allows `VersionInfo` instance, a tuple, or a dict :

```python
def comparator(operator):
    """Wrap a VersionInfo binary op method in a type-check."""

    @wraps(operator)
    def wrapper(self, other):
        comparable_types = (VersionInfo, dict, tuple)  # bug!
        if not isinstance(other, comparable_types):
            raise TypeError(
                "other type %r must be in %r" % (type(other), comparable_types)
            )
        return operator(self, other)

    return wrapper
```

We should add a list and a string to the allowed types. 

Would that be ok @python-semver/reviewers?

tomschr:
This PR fixes #244 and contains the following changes:

The `VersionInfo.compare` method allows already `VersionInfo`, dict, tuple/list and str.

* Extend list of allowed types in `comparator` with list and str.
* Rewrite case `test_should_not_allow_to_compare_version_with_string` ->
  `test_should_compare_version_string`
* Add new test cases:
  - `test_should_compare_version_list`
  - `test_should_not_allow_to_compare_invalid_versionstring`
* Use `pytest.mark.parametrize` for `test_should_compare_version_tuples`
* Update `usage.rst` to document all possibilities
* Update `CHANGELOG`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,7 @@ Bug Fixes
 
 * :gh:`224` (:pr:`226`): In ``setup.py``, replaced in class ``clean``,
   ``super(CleanCommand, self).run()`` with ``CleanCommand.run(self)``
+* :gh:`244` (:pr:`245`): Allow comparison with ``VersionInfo``, tuple/list, dict, and string.
 
 
 Additions
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -295,7 +295,7 @@ To compare two versions depends on your type:
   The return value is negative if ``version1 < version2``, zero if
   ``version1 == version2`` and strictly positive if ``version1 > version2``.
 
-* **Two** :class:`semver.VersionInfo` **types**
+* **Two** :class:`semver.VersionInfo` **instances**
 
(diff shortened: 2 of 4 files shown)


tomschr:
@scls19fr any concerns with this PR? If not, I'd merge it on Saturday to prepare for the release.

Issue #176, opened 2019-11-03
Prepare semver 3.0.0


tomschr:
## Situation

As we have released 2.9.0 of semver, we should consider of moving to 3.0.0 soon. This is needed to start working on adding new features and optimizing source code for Python 3.

(I've created this issue to make the upcoming 3.0.0 more visible and to have a record when we need to search the history.)

## Suggested Solution

In order to create semver 3.0.0, we need to think about the following tasks:

* [x] Create a maintenance branch for 2.x.y line: `maint/v2`
* [x] Remove test target `py27` and `py34` from `tox.ini`
* [x] Tag the 2.9.0 release.
* [x] Update the Trove classifiers in `setup.py` and remove support for Python 2.7 and 3.4.
* [x] Optimize source code for Python 3 only and remove any code that was introduced due to 2.7.
* [x] Provide (and label it correctly) the 2.9.0 documentation for ReadTheDocs.
   We need a configuration that should be possible that old (2.9.x) and new versions (3.*) are still available and can coexit together peacefully. :wink: 
* [ ] Change version number in master to 3.0.0.

Anything else? :wink: 

I've updated the description in https://github.com/k-bx/python-semver/issues/161#issuecomment-539045512 of issue #161

scls19fr:
> Tag the 2.9.0 release

I think it's ever the case... a 2.9.0 release [have ever been tagged](https://github.com/k-bx/python-semver/releases/tag/2.9.0)

tomschr:
> > Tag the 2.9.0 release
> 
> I think it's ever the case... a 2.9.0 release [have ever been tagged](https://github.com/k-bx/python-semver/releases/tag/2.9.0)


Yes, I just mention it to not forget about it. :wink: I've checked it now.

ppkt:
a quick question - do we have any branch for 3.0.0 release?

tomschr:
@ppkt as far as I know, not now. I have some ideas in my mind which I would like to apply to a 3.0.0 only release. :wink: 

My impression was to wait(?) until Python 2.7 is EOL in January 2020. Maybe I got it wrong.

Well, anyway, it should be fairly easy to create a new branch. :wink:

tomschr:
Checked "Provide (and label it correctly) the 2.9.0 documentation for ReadTheDocs."

tlaferriere:
[...]

PR #247, opened 2020-05-05
Prepare semver 3

tomschr:
First attempt to prepare for #176 

Generally, remove everything related to Python 2.7 & 3.4 and add missing 3.8 & 3.9:

* Raise version to indicate that it's different from 2.10.0
* Prepare (empty) CHANGELOG for ~2.99.99~ 3.0.0-dev.1
* Correct Trove categories and python_requires in setup.py
* Adapt `tox.ini` & `.travis.yml`: remove old stuff, add missing (see above)
* Adapt `semver.py`:
  - Remove if clause for cmp
  - Don't check for `__name__`, use `__qualname__` instead
  - Use `yield from ...` (for loop is not necessary anymore)

----

@python-semver/reviewers As 2.10.0 is released now, this is the first attempt to keep the ball rolling toward semver 3! :smile: 

It's a small change, but when merging it to master, we open the development for semver3. All changes will be on top of this.

**Question**: I used version 3.0.0-dev1 for this, ok?

Let me know what you think about. Thanks.


tlaferriere:
I may not be an official reviewer, but I am excited about this.

> **Question:** I used version 3.0.0-dev1 for this, ok?

Sounds good to me

Since this now supports python >=3.5, it now supports type annotations (on functions only until 3.6). I thought this may be a good moment to start adding type hints to the functions you touch, unless you wanted to type hint everything at once in a single PR. I don't know if it would be the right moment to start thinking about #213 .

Anyway, IMHO this is clean and good to go, but that's just me saying that.

tomschr:
@tlaferriere Thanks for your feedback. :+1:  It doesn't matter if you are an "official reviewer", everybody is welcome. :smiley: 

> Since this now supports python >=3.5, it now supports type annotations (on functions only until 3.6). I thought this may be a good moment to start adding type hints to the functions you touch, unless you wanted to type hint everything at once in a single PR. I don't know if it would be the right moment to start thinking about #213 .

Yes, my idea was to introduce this when we "officially" work on version 3 of semver. I have another (local) branch where I did some experiments with it.

Or do you think this is something that you would introduce into the semver 2 line?

tomschr:
I close this issue in favor of #290; integrated the changes here already in the other PR.

PR #250, opened 2020-05-11
Simplify openSUSE installation with obs://

tomschr:
Use `obs://` scheme instead of the longer `http://`. The former has the additional advantage, that the current release is automatically detected.

@m-rey: Would that be ok?


tomschr:
I used the latter (with the underscores). It should be consistent now. Thanks! :+1:

PR #253, opened 2020-05-13
Fix #252: Correct wrong str type in docstring

### LINKED ISSUE #252: bump_build() and bump_prerelease() have the wrong rtype
These two methods return `VersionInfo` objects, the `rtype` in the docstring says that they return a `str`.

tomschr:
Type should be VersionInfo instead of str in ``VersionInfo.bump_build`` and `VersionInfo.bump_prerelease`

@tlaferriere: could you have a look please?

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -328,7 +328,7 @@ class VersionInfo(object):
         untouched.
 
         :return: new object with the raised major part
-        :rtype: VersionInfo
+        :rtype: :class:`VersionInfo`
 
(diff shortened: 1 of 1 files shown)


tlaferriere:
Is the `:class:` prefix necessary? The other bumps use simply `VersionInfo`

tomschr:
Well, Sphinx creates a reference to the class so the user can click the link. IMHO, it's improves usability of the API a bit.

If the others does not contain that prefix, then it's inconsistent. In that case, I'd change that too.

PR #254, opened 2020-05-13
Fix #251: Correct return type for next_version()

### LINKED ISSUE #251: VersionInfo.next_version() returns Union[VersionInfo, str]
`VersionInfo.next_version()` return type is uncertain because at line 463 of semver.py `return version.replace(prerelease=None, build=None)` returns a VersionInfo object, whereas in line 466 it is explicitly converted to `str` and at line 470, `version.bump_prerelease()` is documented with `rtype: str`.  
This is problematic because if you call this on a version that has a prerelease or build number, you get a `VersionInfo` object, and otherwise you get a `str`, and these two types must be handled quite differently.

Personally, I would expect `VersionInfo.next_version()` to return a `VersionInfo` object in all cases, but I think the main problem here is the ambiguity of the return type.

tlaferriere:
@tomschr if you could take a look that would be appreciated.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Features
 Bug Fixes
 ---------
 
-* :gh:`251` (:pr:`000`): Fixed return type of ``semver.VersionInfo.next_version``
+* :gh:`251` (:pr:`251`): Fixed return type of ``semver.VersionInfo.next_version``
   to always return a ``VersionInfo`` instance.
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Features
 Bug Fixes
 ---------
 
-* :gh:`251` (:pr:`251`): Fixed return type of ``semver.VersionInfo.next_version``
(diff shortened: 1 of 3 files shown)


tlaferriere:
PS: you might want to squash these commits since I merged newer changes from the repo into my old fork

tomschr:
@tlaferriere  haha, we had the same idea. :smile:  I've created #255. 

There is one thing you could improve: Could you copy the CHANGELOG from my PR? You should also adapt the PR number to this. Other than that, it looks ok.

tlaferriere:
I was just about to do that

CHANGELOG.rst
```diff
@@ -6,6 +6,32 @@ Change Log
...
+
+
+Features
+--------
+
+
+Bug Fixes
+---------
+
+* :gh:`251` (:pr:`251`): Fixed return type of ``semver.VersionInfo.next_version``
```
tomschr, line 23: The issue number is 251, but this PR number is 254, so I would propose:

```suggestion
* :gh:`251` (:pr:`254`): Fixed return type of ``semver.VersionInfo.next_version``
```

PR #255, opened 2020-05-13
Fix #251: return VersionInfo for next_version

### LINKED ISSUE #251: VersionInfo.next_version() returns Union[VersionInfo, str]

tomschr:
This PR fixes #251 and contains the following changes:

* Return `VersionInfo` instance in `VersionInfo.next_version` and not str when part is major, minor, or patch.
* Change `test_next_version_with_versioninfo`: Test for correct return type (should be `VersionInfo`)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,34 @@ Change Log
 All notable changes to this code base will be documented in this file,
 in every released version.
 
+
+Version 2.10.1 (WIP)
+====================
+
+:Released: 2020-0x-0y
+:Maintainer:
+
+
+Features
+--------
+
+
+Bug Fixes
+---------
+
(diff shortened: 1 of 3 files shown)


tomschr:
Closed in favor of #253

PR #261, opened 2020-06-09
Fix #260 __getitem__ returning `None` on falsy parts

### LINKED ISSUE #260: __getitem__ returns `None` for version parts that are falsy
For example, VersionInfo(0, 1, 0, 'r.1')[2] returns None when it should return 0.

tlaferriere:
For reference [here](https://stackoverflow.com/questions/38339496/python-filternone-list-of-bools-behavior) is the behavior that caused the problem in the first.

I also found that the IndexError wasn't being raised every time it should because it was reliant on the faulty behavior descibed above for some edge cases.


tlaferriere:
Yet again I have not cleaned up my fork, so please squash commits as you merge 😉

semver.py
```diff
@@ -544,6 +544,8 @@ def __getitem__(self, index):
         (3, 4, 5)
         """
         if isinstance(index, int):
+            if index < 0:
+                raise IndexError("Version index cannot be negative")
```
tomschr, line 548: I tried it without these lines and it worked. Therefor, I would remove them:

```suggestion
```

semver.py
```diff
@@ -554,11 +556,14 @@ def __getitem__(self, index):
             raise IndexError("Version index cannot be negative")
 
         # Could raise IndexError:
```
tomschr, line 558: I think, this comment can be removed:

```suggestion
```

tomschr:
I've played around a bit and it seems, I could check for negative indices in the if clause around line 549:

```python
        if (
            isinstance(index, slice)
            and (index.start is None or index.start < 0)
            and (index.stop is None or index.stop <= 0)
            and (index.start < 0 and index.stop <=0)
        ):
            raise IndexError("Version index cannot be negative")
```

this makes your first if clause obsolete.

Furthermore, I think the test cases doesn't distinguish enough between the error messages. Maybe we should rework them, I would suggest these additional changes:

```python
@pytest.mark.parametrize(
    "version, index",
    [
        ("1.2.3", 3),
        ("1.2.3", slice(3, 4)),
        ("1.2.3", 4),
        ("1.2.3", slice(4, 5)),
        ("1.2.3", 5),
        ("1.2.3", slice(5, 6)),
        ("1.2.3-rc.0", 5),
        ("1.2.3-rc.0", slice(5, 6)),
        ("1.2.3-rc.0", 6),
        ("1.2.3-rc.0", slice(6, 7)),
    ],
)
def test_version_info_should_throw_index_error(version, index):
    version_info = VersionInfo.parse(version)
    with pytest.raises(IndexError, match=r"Version part undefined"):
        version_info[index]

@pytest.mark.parametrize(
    "version, index",
    [
        ("1.2.3", -1),
        ("1.2.3", -2),
    ],
)
def test_version_info_should_throw_index_error_when_negative_index(version, index):
    version_info = VersionInfo.parse(version)
    with pytest.raises(IndexError, match=r"Version index cannot be negative"):
        version_info[index]

```

tlaferriere:
I added some cases to `test_version_info_should_throw_index_error_when_negative_index`:
```
        ("1.2.3", slice(-2, 2)),
        ("1.2.3", slice(2, -2)),
        ("1.2.3", slice(-2, -2)),
```
and I noticed that it gave us the wrong error (`"Version part undefined"`), so I changed the second if logic and now it works perfectly.

tomschr:
@tlaferriere This is great, thanks for your efforts :+1:  I forget to tell you that we also need to update the `CHANGELOG.rst` file. Could you amend it?
To make it easier for you, here is the text snippet to copy and paste (add it before the `Version 2.10.1` line):

```rst
Version 2.10.2 (WIP)
==============

:Released: 2020-xx-yy
:Maintainer:

Features
--------

n/a

Bug Fixes
---------

:gh:`260` (:pr:`261`): Fixed ``__getitem__`` returning None on wrong parts


Additions
---------

n/a

Removals
--------

n/a
```

semver.py
```diff
@@ -548,17 +548,19 @@ def __getitem__(self, index):
...
+        part = tuple(filter(lambda p: p is not None, self.to_tuple()[index]))
 
         if len(part) == 1:
             part = part[0]
-        if not part:
+        elif len(part) == 0:
+            raise IndexError("Version part undefined")
+
+        if part is None:
             raise IndexError("Version part undefined")
```
tomschr, line 564: I don't like that we need to raise IndexError two times. Could we combine it? Like this:

```suggestion
        elif part is None or len(part) == 0:
            raise IndexError("Version part undefined")
```

It works for me, can you confirm?

tlaferriere: That does work fine. I'm going to apply it.

tomschr:
Thank you @tlaferriere for your efforts! :heart:  I've squashed the commits and merged it to master.

May I give you a small hint about keeping commits clean? :wink: It looks like you haven't synced your fork with the latest changes from upstream. Find more information in the GitHub documentation [Keep your fork synced](https://help.github.com/en/github/getting-started-with-github/fork-a-repo#keep-your-fork-synced).

PR #264, opened 2020-06-11
Deprecate `max_ver` and `min_ver`

### LINKED ISSUE #160: Deprecate min_ver/max_ver and use builtin max()/min() functions

tlaferriere:
This could close #160 .

As I said in this [comment](https://github.com/python-semver/python-semver/issues/160#issuecomment-642328641), this PR will deprecate functions that could be useful as a shortcut.
As you can see in the new examples in the documentation, it is quite a mouthful to find the maximum of two version strings.

As an alternative to this we could keep `max_ver` and `min_ver` as utility functions that automatically convert from whatever supported types we have, passes them to `max` or `min` then converts back into the input types to return them.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -520,8 +520,7 @@ And the same can be done with tuples:
 
 For dictionaries, it is very similar to finding the max version tuple: see :ref:`_sec.convert.versions`.
 
-The Old Way
-^^^^^^^^^^^
+The "old way" with :func:`semver.max_ver` or :func:`semver.min_ver` is still available, but not recommended:
 
 .. code-block:: python
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -617,6 +617,28 @@ them with code which is compatible for future versions:
      >>> s1 == s2
      True
 
+* :func:`semver.max_ver`
+
+  Replace it with ``max(version1, version2, ...)`` or ``max([version1, version2, ...])``:
+
+  .. code-block:: python
+
+     >>> s1 = semver.max_ver("1.2.3", "1.2.4")
+     >>> s2 = str(max(map(semver.VersionInfo.parse, ("1.2.3", "1.2.4"))))
+     >>> s1 == s2
+     True
+
+* :func:`semver.min_ver`
+
+  Replace it with ``min(version1, version2, ...)`` or ``min([version1, version2, ...])``:
+
+  .. code-block:: python
+
+     >>> s1 = semver.min_ver("1.2.3", "1.2.4")
+     >>> s2 = str(min(map(semver.VersionInfo.parse, ("1.2.3", "1.2.4"))))
+     >>> s1 == s2
+     True
+
 * :func:`semver.parse`
 
   Replace it with :func:`semver.VersionInfo.parse` and
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -483,7 +483,7 @@ That gives you the following possibilities to express your condition:
 .. _sec_max_min:
 
 Getting Minimum and Maximum of Multiple Versions
--------------------------------------------
+------------------------------------------------
 .. versionchanged:: 2.10.2
    The functions :func:`semver.max_ver` and :func:`semver.min_ver` are deprecated in
    favor of their builtin counterparts :func:`max` and :func:`min`.
@@ -492,17 +492,43 @@ Since :class:`semver.VersionInfo` implements :func:`__gt__()` and :func:`__lt__(
 
 .. code-block:: python
 
-    >>> str(max(semver.VersionInfo.parse("1.0.0"), semver.VersionInfo.parse("2.0.0")))
-    '2.0.0'
-    >>> str(min(semver.VersionInfo.parse("1.0.0"), semver.VersionInfo.parse("2.0.0")))
-    '1.0.0'
-
-.. code-block:: python
-
     >>> max([semver.VersionInfo(0, 1, 0), semver.VersionInfo(0, 2, 0), semver.VersionInfo(0, 1, 3)])
     VersionInfo(major=0, minor=2, patch=0, prerelease=None, build=None)
     >>> min([semver.VersionInfo(0, 1, 0), semver.VersionInfo(0, 2, 0), semver.VersionInfo(0, 1, 3)])
     VersionInfo(major=0, minor=1, patch=0, prerelease=None, build=None)
+
+Incidentally, using :func:`map`, you can get the min or max version of any number of versions of the same type
+(convertible to :class:`semver.VersionInfo`).
+
+For example, here are the maximum and minimum versions of a list of version strings:
+
+.. code-block:: python
+
+    >>> str(max(map(semver.VersionInfo.parse, ['1.1.0', '1.2.0', '2.1.0', '0.5.10', '0.4.99'])))
+    '2.1.0'
+    >>> str(min(map(semver.VersionInfo.parse, ['1.1.0', '1.2.0', '2.1.0', '0.5.10', '0.4.99'])))
+    '0.4.99'
+
+And the same can be done with tuples:
+
+.. code-block:: python
+
+    >>> max(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)]))).to_tuple()
+    (2, 1, 0)
+    >>> min(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)]))).to_tuple()
+    (0, 4, 99)
+
+For dictionaries, it is very similar to finding the max version tuple: see :ref:`_sec.convert.versions`.
+
+The Old Way
+^^^^^^^^^^^
+
+.. code-block:: python
+
+    >>> semver.max_ver("1.0.0", "2.0.0")
+    '2.0.0'
+    >>> semver.min_ver("1.0.0", "2.0.0")
+    '1.0.0'
 
 
 Dealing with Invalid Versions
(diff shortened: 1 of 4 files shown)


docs/usage.rst
```diff
@@ -445,7 +445,11 @@ To compare two versions depends on your type:
 
 Other types cannot be compared.
 
-If you need to convert some types into other, refer to :ref:`sec.convert.versions`.
+If you need to convert some types into others, refer to :ref:`sec.convert.versions`.
+
+The use of these comparison operators also implies that you can also use builtin
+functions that leverage this capability; builtins including but not limited to: :func:`max`, :func:`min`
+(refer to :ref:`sec_max_min` for examples) and :func:`sorted`.
```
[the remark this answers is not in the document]
tlaferriere: From what I remember in my French classes, semicolons are used just like a period (to end a sentence), but highlight the link between the two different sentences. I believe it to be relatively similar in English.
I agree that the comma should be there, this sentence lacked pacing
```suggestion
functions that leverage this capability; builtins including, but not limited to, : :func:`max`, :func:`min`
(for examples, see :ref:`sec_max_min`) and :func:`sorted`.
```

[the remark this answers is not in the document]
tomschr: I'm not a native speaker so I can only cite what my English proofeader told me. :wink:  From what I've heard, the use of semicolons *looks* simple, but it actually isn't. But I'm not an expert, so that's why I wrote that I'm not sure.

I don't want to force that upon you, it was just meant as an idea. You can only apply the commas and leave the other as is, if you like. Or rewrite the sentence. :wink:

tomschr:
I forget to add another wish: could you add the two functions into the "Replacing Deprecated Functions" section? See file `docs/usage.rst`, line 542. Thanks! :+1:

tlaferriere:
@tomschr before we go down this rabbit hole any more than we should, can you address [this](https://github.com/python-semver/python-semver/issues/160#issuecomment-643483135) please?

docs/usage.rst
```diff
@@ -476,9 +480,48 @@ That gives you the following possibilities to express your condition:
...
+    '0.4.99'
+
+And the same can be done with tuples:
+
+.. code-block:: python
+
+    >>> max(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)]))).to_tuple()
+    (2, 1, 0)
+    >>> min(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)]))).to_tuple()
+    (0, 4, 99)
```
tomschr, line 519: Not sure if you are aware of that, but some time ago I've configured doctests to be called and tested when running the test suite. This ensures that the released documentation is always accurate and works. Typos or invalid code snippets are a thing of the past. :wink: 

With the current implementation in this branch, the following change works for me:

```suggestion
    >>> max(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)])).to_tuple()
    (2, 1, 0, None, None)
    >>> min(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)])).to_tuple()
    (0, 4, 99, None, None)
```

tomschr:
@tlaferriere Could you add the issue and pr number in the `CHANGELOG.rst` file? I'd suggest to add the following line to the `Removals` section:

```rst
* :gh:`160` (:pr:`264`): Output DeprecationWarning for ``semver.max_ver`` and ``semver.min_ver``.
```

docs/usage.rst
```diff
@@ -476,9 +480,48 @@ That gives you the following possibilities to express your condition:
...
+    (2, 1, 0, None, None)
+    >>> min(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)])).to_tuple()
+    (0, 4, 99, None, None)
+
+For dictionaries, it is very similar to finding the max version tuple: see :ref:`_sec.convert.versions`.
 
-Getting Minimum and Maximum of two Versions
--------------------------------------------
+The Old Way
+^^^^^^^^^^^
```
tomschr, line 524: From a documentation strategy, it's better to avoid "lonely sections" (=only one subsection inside a parent section). Lonely section doesn't add value as long as you don't add another subsection.

Therefor, I would suggest to remove this and add maybe a small sentence like:

```suggestion
The "old way" with :func:`semver.max_ver` or :func:`semver.min_ver` is still available, but not recommended:
```

(Feel free to rephrase it if you find a better way, was just an example.)

tlaferriere:
> Could you add the issue and pr number in the `CHANGELOG.rst` file? I'd suggest to add the following line to the `Removals` section:

Sure!

While we're at it, I find that the name `Removals` might be a bit misleading since this is not a major release (yet 😉). Maybe renaming it to `Deprecations` would be nicer? It would also increase the importance of this section once we do remove things in version 3.0 (or 3.1).
Also it would make it DRYer since all we would have to add now is the issue and PR number and the name of the deprecated functions.

tomschr:
> While we're at it, I find that the name `Removals` might be a bit misleading since this is not a major release (yet ). Maybe renaming it to `Deprecations` would be nicer?

You can read my mind! :laughing: After I've wrote my message, I thought the same but didn't change it.

So, yes please! Go ahead and use Deprecations. :+1:

tlaferriere:
I also went ahead and removed empty sections in the changelog, figured it would look cleaner like this.

tomschr:
Thanks you Thomas. Much appreciated! :+1: 

As I've mentioned in https://github.com/python-semver/python-semver/pull/264#issuecomment-643477924, we should add the two deprecated functions also in the section [Replacing Deprecated Functions](https://python-semver.readthedocs.io/en/latest/usage.html#replacing-deprecated-functions). The list is sorted, so I'd suggest to add it between `semver.format_version` and `semver.parse`. Ok?

I think, that will complete it and we can merge it to master. :wink:

tomschr:
Oh, by the way: I plan to release 2.10.2 when this is merged.

docs/usage.rst
```diff
@@ -476,9 +480,47 @@ That gives you the following possibilities to express your condition:
...
+.. code-block:: python
+
+    >>> max(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)])).to_tuple()
+    (2, 1, 0, None, None)
+    >>> min(map(lambda v: semver.VersionInfo(*v), [(1, 1, 0), (1, 2, 0), (2, 1, 0), (0, 5, 10), (0, 4, 99)])).to_tuple()
+    (0, 4, 99, None, None)
 
-Getting Minimum and Maximum of two Versions
--------------------------------------------
+For dictionaries, it is very similar to finding the max version tuple: see :ref:`_sec.convert.versions`.
```
tomschr, line 521: I've built the documentation and this ID is unknown (due to the underscore). It's a typo. If you remove it, it works:

```suggestion
For dictionaries, it is very similar to finding the max version tuple: see :ref:`sec.convert.versions`.
```

tlaferriere:
I'm out for a few hours, if there are any more typos go ahead and fix them so you can merge this soon.

tomschr:
@tlaferriere! I think, it's in a state now that we can merge it to master. Haven't found any further issues.

Thank you very much for all your efforts! :+1:  Will merge it.

PR #265, opened 2020-06-11
Add type hints

### LINKED ISSUE #213: Consider type hints for 3.x.y release?

tlaferriere:
This adds type hints to every function not deprecated. Should fix #213 .

I've had to refactor some of the code to get mypy to stop complaining on python 3.8. It isn't configured in Tox yet because I am waiting on #247 to be merged to avoid conflicts.

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -50,8 +50,8 @@ __all__ = (
 
 from types import FrameType
 
-from typing import Callable, TypeVar, SupportsInt, Protocol, Optional, Tuple, Dict, Union, Generator, cast, Iterable, \
-    List, AnyStr
+from typing import Callable, TypeVar, SupportsInt, Optional, Tuple, Dict, Union, Generator, cast, Iterable, \
+    List
 
 #: Contains the implemented semver.org version of the spec
 SEMVER_SPEC_VERSION = "2.0.0"
@@ -167,11 +167,6 @@ def comparator(operator: Comparator) -> Comparator:
     return wrapper
 
 
-class SupportsString(Protocol):
-    def __str__(self):
-        ...
-
-
 class VersionInfo:
     """
     A semver compatible version class.
@@ -210,7 +205,7 @@ class VersionInfo:
     )
 
     def __init__(self, major: SupportsInt, minor: SupportsInt = 0, patch: SupportsInt = 0,
-                 prerelease: SupportsString = None, build: SupportsString = None):
+                 prerelease: Union[str, int] = None, build: Union[str, int] = None):
         self._major = int(major)
         self._minor = int(minor)
         self._patch = int(patch)
--- a/semver.py
+++ b/semver.py
@@ -3,12 +3,11 @@ from __future__ import print_function
 
 import argparse
 import collections
-from functools import wraps, partial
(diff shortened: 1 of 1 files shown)


tomschr:
@tlaferriere Thanks Thomas for investigating that. As I did lots of changes lately with the upcoming 3.0.0-alpha0, I'd suggest to add your work into one of the next alphas or betas.

I think, your work and effort should be preserved. :+1: We only need to make sure we fix all the conflicts. :wink:

tlaferriere:
Is there a branch for 3.x development I can merge this in?

tomschr:
You could use the `feature/3.0.0-alpha0` branch. However, maybe we should try this with alpha1 to make the changes a bit more manageable. Would that be an option for you?

tlaferriere:
Alright I'll point a pr to the Alpha 0 branch so I can start fixing the conflicts.

tomschr:
That would be awesome! Thanks and sorry for all the conflicts.

tlaferriere:
Closing in favor of tomschr/python-semver#6

Issue #266, opened 2020-06-11
Make __getitem__ support strings as dict lookup 


tlaferriere:
# Use Case
After using number indices on `VersionInfo`, I intuitively used a string index.
I believe it is the next logical step in the `__getitem__` feature line.

# Situation
We can use a numeric index to access the tuple representation of `VersionInfo` implicitly, but a string index (meant to implicitly access the dict representation) fails because it is simply not implemented

# Proposition
When `__getitem__` receives a string, use it as index to the dict representation.

tomschr:
@tlaferriere Thanks for the idea!

So basically you mean allowing this code:

```python
>>> ver = semver.VersionInfo.parse("1.2.3")
>>> ver[1] == ver["minor"]
True
```

Hmn... :thinking: ... why would anyone do that if a developer can simply write `ver.minor`? Why is `ver["minor"]` "better" than `ver.minor`?

Apart from this, the Zen of Python states "there should be one---and **preferably only one**---obvious way to do it." (emphasize by me) I know, it's not a strict law, but a system/recommendation/philosophy to code.

This sounds as an indication that we would create two competing ways to do it. Technically it is easy to implement. Not sure whether we really should do that.

Maybe you have a really convincing argument that I overlooked?

What do @python-semver/reviewers think?

tlaferriere:
@tomschr you're right.
I came across this in my code where I was using an `IntEnum` to access the parts programmatically using indices, but then I had to do `update_type.name.lower()` to use the `VersionInfo.next_version` method since it only takes strings (which makes sense) to describe the targeted part. This somehow made an impression on me because later I tried to use `update_type.name.lower()` inadvertently to access a version part through `__getitem__`, which failed for obvious reasons, to which I remedied (inadvertently yet again 🤦 ) by indexing `VersionInfo.to_dict`.
In conclusion, that's what `getattr` is for (getting something out of an object using a string), therefore this is redundant and goes against the zen of python.

P.S. Probably should have slept more the night before 😜

tomschr:
@tlaferriere All good, don't worry. :+1:  Haha, I know this too, when I'm caught by some idea. :stuck_out_tongue_winking_eye: 

That's why we discuss these things, raise questions and come to a conclusion. Thanks for your ideas anyway, I appreciated it very much. Every idea can improve this little project. :+1:

PR #267, opened 2020-06-11
Add string support to __getitem__

tlaferriere:
Pretty simple implementation for #266 , fully tested.

I also took the initiative to increase coverage a bit.


tlaferriere:
I covered everything except what was left in  `max_ver` and `min_ver` because they might be deprecated soon.

PR #268, opened 2020-06-12
Increase coverage

tlaferriere:
Tried to cover everything but max_ver and min_ver since they might be deprecated.

I thought it could be nice to start off version 3 with 100% coverage


tomschr:
This is wonderful @tlaferriere, thank you very much! :+1:  I will squash all the commits into one.

PR #273, opened 2020-06-16
Configure and add Towncrier files

### LINKED ISSUE #270: Use Towncrier for Changelog?
# Situation

Currently, all changes are recorded in a single `CHANGELOG.rst` file. In most cases this is enough and it works well.

However, in some cases conflicts can occur. Sometimes it would be helpful to have a tool which could create that changelog automatically.

# Idea

Some projects use the [towncrier](https://pypi.org/project/towncrier/) tool. It is built around "news fragments", little text snippets that are stored in different files under a common directory (e.g. `changes.d`). When a new release is drafted, these files are brought together and create the final changelog file. This minimizes any conflicts.

Every filename contains an issue number and a "type". A type can be feature, bugfix, doc, removal, or you can use self-defined types. These types will become sections in the final changelog file. All fragments are collected, sorted, and added under the sections.

When Towncrier is correctly set up, a typical workflow looks like this:

1. Start working on a specific issue, let's say, `123`.
1. Decide what type this issue is, for example, feature, bugfix, etc.
1. Combine the two information and create a fragment file: `towncrier create 123.bugfix`.
    This file will be added inside the `changes.d` directory.
1. Describe the issue.
1. Commit the file.
1. Merge issue to master.

The more issues you solve, the more files you will collect inside your `changes.d` directory. At some point, we need to create a new release. In this case, the workflow looks like this:

1. Create a new release branch.
   (could be done in other ways too, but this makes things a bit easier).
1. Raise the version.
1. Produce a draft changelog to see everything is included with: `towncrier build --draft`.
1. If satisfied, produce the final changelog file with: `towncrier build`.
    This command will remove all files inside the `changes.d` directory and create the final changelog file.
1. Commit all changes and merge it.

## Pros

* All changes reside in single files under a common directory, for example, `changes.d`.
  This makes it easier to see which will end up in the new release.
* Reduces conflicts as changes are stored in separate files. 
* Able to configure 

## Cons

* Works on Python 3.5+ only.
  (This may be not as a big disadvantage as we move to Python 3 soon.)
* Every file has to be named with ISSUE-TYPE. This may not always appropriate.
* Towncrier works best in a development system where all merges involve closing a ticket.

@python-semver/reviewers @tlaferriere  what do you think about that? Would that be helpful or just too much?

tomschr:
This PR fixes #270 and contains the following changes:

* Add `changelog.d/.gitignore` to keep this directory
* Create `changelog.d/README.rst` with some descriptions
* Add `changelog.d/_template.rst` as Towncrier template
* Add `[tool.towncrier]` section in `pyproject.toml`
* Add "changelog" target into `tox.ini`. Use it like `tox -e changelog -- CMD` whereas CMD is a towncrier command. The default `tox -e changelog` calls towncrier to create a draft of the changelog file and output it to stdout.
* Update documentation and add include a new section "Changelog" included from `changelog.d/README.rst`

----

@python-semver/reviewers @tlaferriere  this PR is an attempt to show how this could work. 

The `changelog.d` directory contains some example files whereas only file `270.feature.rst` is "real", the other newsfragments will be removed before merging to master.

I've documented the procedure, see `changelog.d/README.rst`. If you prefer to read it as HTML, build the documentation with `tox -e docs` and look into "Contributing to semver" / "Changelog".

Would like to hear your ideas, feedback, or comments. Thank you!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,20 @@ Change Log
 All notable changes to this code base will be documented in this file,
 in every released version.
 
+Changes for the upcoming release can be found in
+the `"changelog.d" directory <https://github.com/python-semver/python-semver/tree/master/changelog.d>`_
+in our repository.
+
+..
+   Do *NOT* add changelog entries here!
+
+   This changelog is managed by towncrier and is compiled at release time.
+
+   See https://python-semver.rtd.io/en/latest/development.html#changelog
+   for details.
+
+.. towncrier release notes start
+
 
 Version 2.10.2
 ==============
@@ -32,11 +46,15 @@ Additions
 n/a
 
 Deprecations
---------
+------------
 
 * :gh:`160` (:pr:`264`):
     * :func:`semver.max_ver`
     * :func:`semver.min_ver`
+
+
+----
+
 
 
 Version 2.10.1
@@ -60,6 +78,9 @@ Bug Fixes
 
 * :gh:`251` (:pr:`254`): Fixed return type of ``semver.VersionInfo.next_version``
   to always return a ``VersionInfo`` instance.
+
+
+----
 
 
 Version 2.10.0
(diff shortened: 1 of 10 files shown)


tomschr:
This will be integrated into PR #290 so closing this.

PR #275, opened 2020-07-10
Fix #274:  String Types Py2 vs. Py3 compatibility

### LINKED ISSUE #274: Py2 vs. Py3 incompatibility: TypeError
# Situation

I'm confused by the statement in the Readme that seems to say 2.10.0 _does_ still support Python 2.7. That's not what I'm seeing. I have a project that uses this package and runs in Py2, and as soon as it started using the 2.10 release, I started getting errors like this which look to me like what you'd get from a Py2 vs. Py3 incompatibility:

    TypeError("Expected str or VersionInfo instance, but got <type 'unicode'>"

When I pin the dependency to `<2.10`, the error goes away.

_Originally posted by @eli-darkly in https://github.com/python-semver/python-semver/issues/161#issuecomment-654455352_

tomschr:
This fixes #274 and contains:

fixes problems between different string types. In Python2 `str` vs. `unicode` and in Python3 `str` vs. `bytes`.

---

@eli-darkly I've tried to recycle some code from the six project. Maybe there is an easier way, maybe we can improve some ugly code. Would you like to have a look at? Thank you! :heart:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ Bug Fixes
 ---------
 
 * :gh:`276` (:pr:`277`): VersionInfo.parse should be a class method
+* :gh:`274` (:pr:`275`): Py2 vs. Py3 incompatibility TypeError
 
 
 Additions
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -28,6 +28,7 @@ Significant contributors
 * Carles Barrobés <carles@barrobes.com>
 * Craig Blaszczyk <masterjakul@gmail.com>
 * Damien Nadé <anvil@users.noreply.github.com>
+* Eli Bishop <eli-darkly@users.noreply.github.com>
 * George Sakkis <gsakkis@users.noreply.github.com>
 * Jan Pieter Waagmeester <jieter@jieter.nl>
 * Jelo Agnasin <jelo@icannhas.com>
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -45,10 +45,17 @@ creating a version:
 
 A :class:`semver.VersionInfo` instance can be created in different ways:
 
-* From a string::
+* From a string (a Unicode string in Python 2)::
 
     >>> semver.VersionInfo.parse("3.4.5-pre.2+build.4")
     VersionInfo(major=3, minor=4, patch=5, prerelease='pre.2', build='build.4')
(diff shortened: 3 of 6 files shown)


test_py2.py
```diff
@@ -0,0 +1,102 @@
...
+    assert result == -1
+    result = semver.compare(semver.b("1.1.0"), semver.u("1.2.2"))
+    assert result == -1
+
+
+class TestEnsure:
+    # From six project
+    # grinning face emoji
+    UNICODE_EMOJI = semver.u("\U0001F600")
+    BINARY_EMOJI = b"\xf0\x9f\x98\x80"
```
eli-darkly, line 41: I was thinking that you wouldn't be able to do this in a single file because Python 2.x would not recognize the `b"..."` syntax. But, looking at the docs again (since I'm really not that expert in Python), I see that Py2 doesn't consider `b"..."` to be an error; it just ignores the `b` and treats it as a regular string. So I guess this will work... but I do find it somewhat confusing at first glance, so you might want to add a comment to clarify that this code is being interpreted very differently in Py2 vs. Py3.

eli-darkly: I also find it mildly confusing that this file is called `test_py2.py` even though it gets run in both Python 2 and Python 3. I understand that the point is to verify that it behaves consistently between the two with equivalent inputs, I'm just not sure that that would be clear to someone who was looking at this code without having seen our previous conversation. But then again I'm a little sleepy now and so I find lots of things mildly confusing.

eli-darkly: But I really should emphasize that I'm not super adept at Python. You may want to run this PR by someone else who's had to deal with a lot of Py2/Py3 migration issues because I'm not sure I would catch any real problems.

tomschr: > recognize the `b"..."` syntax. [...] So I guess this will work... but I do find it somewhat confusing at first glance, so you might want to add a comment to clarify that this code is being interpreted very differently in Py2 vs. Py3.

Yes, I've got it from the six package which deals with Py2 vs. Py3 compatibility a lot. :wink: 


> I also find it mildly confusing that this file is called `test_py2.py`

Yes, I will renamed it to `test_typeerror-274.py`. It's probably not ideal either, but it gives a reference to this issue.

tomschr:
@eli-darkly Eli, sorry for the long delay. I've added you as a contributor and documented this in our documentation. If you would like to have a final look I can merge and create a new release.

eli-darkly (approved):

Issue #276, opened 2020-07-21
VersionInfo.parse should be a class method


Anvil:
I wanted to create a subclass of VersionInfo for company reasons (more checkings, more methods, etc) and it turned out the VersionInfo.parse is a staticmethod, always returning VersionInfo instances, no matter what.

By turning the `parse` static method into a class method and making it `return cls(**version_parts)` the result of parse will be an instance of the subclass (my class), not the superclass (VersionInfo).

e.g:

```python

class MyBogusVersionInfo(semver.VersionInfo):

    @classmethod
    def parse(cls, version):
        if version.startswith("foobar"):
            version = version[6:]
        return super().parse(version)

print(type(MyBogusVersionInfo.parse("1.2.3")))
```

Current result is : `<class 'semver.VersionInfo'>`, while I would have expected `<class 'MyBogusVersionInfo'>`

tomschr:
Thanks @Anvil for your issue and the pull request. :+1: 

That's an interesting approach. I see your point. However, not entirely sure if changing `.parse` from a staticmethod to a classmethod is completely backward compatible... :thinking: 

Apart from that, the `.parse` method could be removed in semver3. We've discussed this in issue #258. For semver3, an idea is to "overload" the initializer to allow more datatypes. Maybe that could solve your use case. If we go this route, that would make the `.parse` and adapting it kind of obsolete.

Do you need it for Python2 only or do you use it for Python3?

Thanks for your pull request (PR). As the above discussion is not really solved, I would like to wait for this issue until the above issue is remedied.

Just a short comment about your PR: you don't have to change anything right now (due to the open discussion). It's just a reminder for future PRs. We require test cases for each new feature. :wink: If a feature changes the public API, this also needs to be documented. Ideally with some code examples. For further details, refer to our [Contributing to semver](https://python-semver.readthedocs.io/en/latest/development.html) section.

Anvil:
Hello

As far as I can tell, there wont be any breaking by changing static to class. The reference "VersionInfo.parse" makes it automatically bound to the class.

In general, *all* factory methods (methods that return an instance of the class) should be class methods, not staticmethods.

I dont use python 2 anymore. Do you intent do keep compatibility with python2 ?

Anvil:
I can provide tests and examples if it's accepted, though. Feel free to ask.

Anvil:
[...]

PR #277, opened 2020-07-21
Turn VersionInfo.parse into classmethod to allow subclasses

Anvil:
Suggested solution for #276

### DIFF
--- a/semver.py
+++ b/semver.py
@@ -637,8 +637,8 @@ build='build.10')
 
         return cmp_res in possibilities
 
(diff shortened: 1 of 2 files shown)


tomschr:
@tlaferriere Thomas, I think, this could be merged safely, right?

tlaferriere:
Looks good to me 👍

tomschr:
@Anvil  / @tlaferriere Thank you! I've added also a small test case.

PR #283, opened 2020-10-15
Ensure equal versions have equal hashes

sbrudenell:
Per [the python datamodel](https://docs.python.org/3/reference/datamodel.html#object.__hash__), `a == b` implies `hash(a) == hash(b)`. The current implementation of `VersionInfo.__hash__` includes the build, which isn't used for equality.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -452,6 +452,30 @@ functions that leverage this capability; builtins including, but not limited to:
 (for examples, see :ref:`sec_max_min`) and :func:`sorted`.
 
 
+Version Equality
+----------------
+
+Note that when comparing two versions, the ``build`` part is ignored::
+
+    >>> v = semver.VersionInfo.parse("1.2.3-rc4+1e4664d")
+    >>> v == "1.2.3-rc4+dedbeef"
+    True
+
+This also applies when a :class:`semver.VersionInfo` is a member of a set, or a
+dictionary key::
+
+    >>> d = {}
+    >>> v1 = semver.VersionInfo.parse("1.2.3-rc4+1e4664d")
+    >>> v2 = semver.VersionInfo.parse("1.2.3-rc4+dedbeef")
+    >>> d[v1] = 1
+    >>> d[v2]
+    1
+    >>> s = set()
+    >>> s.add(v1)
+    >>> v2 in s
+    True
+
+
 
 Comparing Versions through an Expression
 ----------------------------------------
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -453,7 +453,7 @@ functions that leverage this capability; builtins including, but not limited to:
 
 
 Determining Version Equality
-----------------
+----------------------------
 
 Version equality means for semver, that major, minor, patch, and prerelease
 parts are equal in both versions you compare. The build part is ignored.
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -452,10 +452,12 @@ functions that leverage this capability; builtins including, but not limited to:
 (for examples, see :ref:`sec_max_min`) and :func:`sorted`.
 
 
-Version Equality
+Determining Version Equality
 ----------------
 
-Note that when comparing two versions, the ``build`` part is ignored::
+Version equality means for semver, that major, minor, patch, and prerelease
+parts are equal in both versions you compare. The build part is ignored.
+For example::
 
     >>> v = semver.VersionInfo.parse("1.2.3-rc4+1e4664d")
     >>> v == "1.2.3-rc4+dedbeef"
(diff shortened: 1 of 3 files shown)


tomschr:
@sbrudenell Thanks for this contribution! Much appreciated! :+1: 

I like it. :+1: My only concern is that this change breaks compatibility with previous releases, right? Maybe this change should go into semver 3?

Futhermore, we should document this behavior somewhere in `docs/usage.rst`. Maybe in `Comparing Version` or in a separate section.
It doesn't need much. I would describe a general example and then give the specific one with a build part. That should include a text that the build part isn't taken into account.

Does that make sense to you?

sbrudenell:
IMO it's a stretch to say this change breaks compatibility. Technically true, but any code relying on the current behavior is relying on the python data model being broken, which is pathological enough that I wouldn't bother supporting it. Up to you though.

Is there a separate branch for changes going into 3.x?

I'll update the docs.

tomschr:
> IMO it's a stretch to say this change breaks compatibility. [...]

Yes, maybe my question was not entirely correct. It was more of a "thinking out loud" approach and to get your opinion. Probably I'm too concerned about that, maybe it's not a big issue. :wink: But it's something that I consider first before accepting any PR. So let me think about that.


> Is there a separate branch for changes going into 3.x?

At the moment this is not the case, unfortunately. I hope we can leave semver2 behind soon and start working on semver 3. For you, nothing changes at the moment. 

> I'll update the docs.

That would be great! :+1:

docs/usage.rst
```diff
@@ -452,6 +452,30 @@ functions that leverage this capability; builtins including, but not limited to:
 (for examples, see :ref:`sec_max_min`) and :func:`sorted`.
 
 
+Version Equality
```
tomschr, line 455: Could we have the title parallel to the others? I think, there is a verb missing. So maybe something like this:

```suggestion
Determining Version Equality
```

docs/usage.rst
```diff
@@ -452,6 +452,30 @@ functions that leverage this capability; builtins including, but not limited to:
...
+----------------
+
+Note that when comparing two versions, the ``build`` part is ignored::
```
tomschr, line 458: I think, we need to explain what _version equality_ is or how we understand that. Maybe something like this:

```suggestion
Version equality means for semver, that major, minor, and patch parts are
equal in both versions you compare. The build part is ignored, for example::
```

Maybe you can find a better phrase. :wink:

docs/usage.rst
```diff
@@ -452,6 +452,30 @@ functions that leverage this capability; builtins including, but not limited to:
...
+    >>> v == "1.2.3-rc4+dedbeef"
+    True
+
+This also applies when a :class:`semver.VersionInfo` is a member of a set, or a
+dictionary key::
+
+    >>> d = {}
+    >>> v1 = semver.VersionInfo.parse("1.2.3-rc4+1e4664d")
+    >>> v2 = semver.VersionInfo.parse("1.2.3-rc4+dedbeef")
+    >>> d[v1] = 1
```
tomschr, line 470: Sorry, I'm not sure what are you trying to show here. Probably I've overlooked something. Could you enlighten me? :-)

tomschr (commented):
Thanks @sbrudenell for your additions. Much appreciated.

I have some minor comments, maybe we can work together to solve them. :-)

docs/usage.rst
sbrudenell: I was trying to illustrate that semver-style equality means that if two versions are equal with different `build`s, then python won't be able to distinguish them as distinct dict keys or set members.

This may be helpful or harmful to authors depending what they're trying to do, but in any case it's surprising in Python that `Obj(args) == Obj(different_args)` (except for primitives). There isn't too much precedent for it. So I was trying to think of important implications of this design and illustrate those to reduce the surprise.

tomschr: Ahh, now I got it. Thanks! :+1:

tomschr:
I'm currently thinking if this change could be safely integrated into the current semver2 code stream or better added to semver3.

@tlaferriere @python-semver/reviewers  Thoughts?

tlaferriere:
This does feel like a bugfix that could be integrated in the 2.12.x line. All we have to do is implement it 😉. So much for freezing 2.10.x 😂.
I don't have much spare time atm but I see you're on a roll @tomschr so if you feel like getting that done, it's up to you to decide whether you want to release it as a bugfix or in in the next major change.

tomschr:
I've tried to read the semver spec to find a dedicated section regarding to compare two versions. The closed I could find was [§10](https://semver.org/#spec-item-10):

```
10. [...] Build metadata MUST be ignored when determining version precedence.
Thus two versions that differ only in the build metadata, have the same precedence.
```

I think, this can be interpreted as our use case and making this a bugfix that should be integrated into semver2 and 3.

Thanks @sbrudenell and @tlaferriere for your contributions and help! Much appreciated! :+1:

PR #292, opened 2020-10-18
Fix #291: Disallow negative numbers in VersionInfo

### LINKED ISSUE #291: semver.VersionInfo allows negative numbers
# Situation

If you pass negative numbers into `semver.VersionInfo` it doesn't raise any errors:

```python
>>> import semver
>>> semver.VersionInfo(-2)                                                                                                                           
VersionInfo(major=-2, minor=0, patch=0, prerelease=None, build=None)
```

Tested with latest 2.11.0

# Proposal

I guess, this shouldn't happen. We should raise a ValueError exception.
Would integrate that into a 2.12.0 release and into the upcoming 3.0.0-alpha0.

@tlaferriere Thomas, can you conform? Do you agree?

tomschr:
This PR fixes #291 and contains the following change:

Disallow negative numbers of `major`, `minor`, and `patch` in `semver.VersionInfo.`
Reason: a version can only contain positive numbers

@tlaferriere This is the corresponding pull request. Maybe there are better ways to do it. :wink: 

I would release 2.12.0 if you think this is useful.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,36 @@ Change Log
 All notable changes to this code base will be documented in this file,
 in every released version.
 
+Version 2.12.0
+==============
+
+:Released:
+:Maintainer: Tom Schraitle
+
+Features
+--------
+
+n/a
+
+
+Bug Fixes
+---------
+
+* :gh:`291` (:pr:`292`): Disallow negative numbers of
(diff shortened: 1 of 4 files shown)


tlaferriere:
Is it a feature or a semantic bugfix? This kind of feels like bug. I would kind of expect it to reject negative numbers out of the box.

PR #317, opened 2020-11-06
Fix #303: Fix Version.__init__ method

### LINKED ISSUE #303: Rethinking __init__ of VersionInfo class?
# Situation

_Originally posted by @tomschr in https://github.com/python-semver/python-semver/issues/258#issuecomment-652482631_

From the above discussion, I thought it would be worth to decouple the `compare` discussion from the initializer discussion. Both are somewhat related, but can live independently. 

With a more "advanced" initializer/constructor we get the following benefits:

* more "pythonic": it's one, obvious way to get an instance.
* avoids the longer function call `VersionInfo.parse(...)`.
* more readable

With such an (overloaded?) initializer, we could cover the following use cases:

```python
>>> from semver import VersionInfo as Version
>>> Version(1)
VersionInfo(major=1, minor=0, patch=0, prerelease=None, build=None)
>>> Version(1, "4", b"5")
VersionInfo(major=1, minor=4, patch=5, prerelease=None, build=None)
>>> Version(1, patch=2)
VersionInfo(major=1, minor=0, patch=2, prerelease=None, build=None)
>>> Version("1.2.3")
VersionInfo(major=1, minor=2, patch=3, prerelease=None, build=None)
>>> Version(b"1.2.3")
VersionInfo(major=1, minor=2, patch=3, prerelease=None, build=None)
>>> v = Version(b"2.3.4")
>>> Version(v)
VersionInfo(major=2, minor=3, patch=4, prerelease=None, build=None)
>>> t = (1, 2, 3)
>>> Version(*t)                                             
VersionInfo(major=1, minor=2, patch=3, prerelease=None, build=None)
>>> d = {'major': 3, 'minor': 4, 'patch': 5,  'prerelease': 'pre.2', 'build': 'build.4'}
>>> Version(**d)
VersionInfo(major=3, minor=4, patch=5, prerelease='pre.2', build='build.4')
```

# Discussions and Possible Solutions

To implement a somewhat more advanced constructor/initializer, we have these options:

1. Program it manually with `isinstance` and if...else constructs
1. Use the [`typing.overload`](https://docs.python.org/3.6/library/typing.html#typing.overload) function (suggested by @tlaferriere)
1. Use [`functools.singledispatch`](https://docs.python.org/3.6/library/functools.html#functools.singledispatch)

However, all three comes at a cost or an issue:

1. Maybe not impossible, but the code would look probably ugly.
1. This seem to be more for the type checker than for an actual implementation. Furthermore, the function distinguishes by types (because of its design), not in regards to the amount of arguments.
1. Is not possible with an `__init__` method. The `singledispatch` works only for _functions_(!), *not* methods. For methods we would need [`functools.singledispatchmethod`](https://docs.python.org/3.8/library/functools.html#functools.singledispatchmethod) which is only available from Python >= 3.8.

Thoughts? Any other ideas? Would that be worth the effort?

tomschr:

* Allow different variants to call Version
* Adapt the documentation and README
* Adapt and amend tests
* Add changelog entries

TODO: Fix typing errors.

@tlaferriere Thomas, could you have a look please? Especially the types, they seem to be a bit tricky. :wink:


src/semver/version.py
```diff
@@ -144,27 +162,122 @@ class Version:
...
+        build = cls._ensure_str(build or verlist[4])  # type: ignore
+
         self._major = version_parts["major"]
         self._minor = version_parts["minor"]
         self._patch = version_parts["patch"]
         self._prerelease = None if prerelease is None else str(prerelease)
         self._build = None if build is None else str(build)
 
+    @classmethod
+    def _ensure_str(cls, s: Optional[String], encoding="UTF-8") -> Optional[str]:
```
tlaferriere, line 247:
```suggestion
    def _ensure_str(cls, s: Optional[StringOrInt], encoding="UTF-8") -> Optional[str]:
```
This should resolve type issues you might be having earlier when using that method.

src/semver/version.py
```diff
@@ -144,27 +162,122 @@ class Version:
 
     def __init__(
```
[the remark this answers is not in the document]
tomschr: Ahh, you mean something like this?

```python
from typing import overload

@overload
def __init__(self, ):  # for Version("1.2.3", ...)
    ...

@overload
def __init__(self, ):  # for Version(1, 2, 3)
    ...

@overload
def __init__(self, ):  # for Version(
    # the real implementation
```

Unfortunately, I don't know how these type annotations should look like. Types are new topic for me, but I like the idea. Any hints for me? :wink:

src/semver/version.py
tomschr: Thanks. I had to change the return type as well (to `Optional[StringOrInt]`), otherwise I get the error message:

    src/semver/version.py:228: error: Incompatible return value type (got "Union[str, int, None]", expected "Optional[str]")

src/semver/version.py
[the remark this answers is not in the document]
tlaferriere: Well the overloads simply specify which signatures are valid. For example, if the version is given as a string, assuming the string has all the information we need, we don't have to (and we shouldn't) pass any of the major, minor, patch, etc to the function. In our case, the type hint without overload would look like this: `(version_string_or_major: Union[str, SupportInt], minor: SupportsInt = None, ...)` which is ugly and is totally useless to the user of the function.

Now with overload it would look like this:
```python
from typing import overload

@overload
def __init__(self, version_string: str) -> Version:  # for Version("1.2.3", ...)
    ...

@overload
def __init__(self, major: SupportsInt, minor: SupportsInt, patch: SupportsInt) -> Version:  # for Version(1, 2, 3)
    ...

# We should do something about the kwargs like `prerelease` and `build`.
# Probably best to throw them in the implementation function signature.

# Do not put @overload here, because this is the actual implementation.
def __init__(self, *args, prerelease: String = None, build: String = None, **kwargs) -> Version:  # for Version(
    # the real implementation
```
And now we can call the function in each way:
```python
Version("1.2.4")  # correct
Version(1, 2, 4)  # correct
Version("1.2.4", 3, 2)  # type error (we should probably also raise a NotImplementedError at runtime)
```

Now you could chuck that code in there, but I still have doubts about how mypy will handle the `prerelease` and `build` kwargs in relation to the three main parts of the version. This would require further testing and experimentation. For now, I think it lays the foundation, and we can add more overloads as we find out what we can do correctly with this method.

src/semver/version.py
tlaferriere: Make sure you use `Optional[str]` here, since it is meant to ensure that it is a `str`.

src/semver/version.py
```diff
@@ -92,21 +110,81 @@ class Version:
...
+            )
 
         for name, value in version_parts.items():
             if value < 0:
                 raise ValueError(
                     "{!r} is negative. A version can only be positive.".format(name)
                 )
 
+        prerelease = cls._ensure_str(prerelease or verlist[3])  # type: ignore
+        build = cls._ensure_str(build or verlist[4])  # type: ignore
```
tlaferriere, line 186: Here is the proper way to use `_ensure_str`:
```suggestion
        if isinstance(prerelease, int):
            self._prerelease = prerelease
        else:
            self._prerelease = cls._ensure_str(prerelease or verlist[3])
        if isinstance(build, int):
            self._build= build
        else:
            self._build = cls._ensure_str(build or verlist[4])
```

src/semver/version.py
```diff
@@ -136,6 +214,43 @@ def cmp_prerelease_tag(a, b):
         else:
             return _cmp(len(a), len(b))
 
+    @classmethod
+    def _ensure_str(
+        cls, s: Optional[StringOrInt], encoding="UTF-8"
+    ) -> Optional[StringOrInt]:
```
tlaferriere, line 220:
```suggestion
        cls, s: Optional[String], encoding="UTF-8"
    ) -> Optional[str]:
```

tlaferriere (changes requested):
My bad for leading you astray in the last review, this should be more correct now

src/semver/version.py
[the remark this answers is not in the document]
tomschr: Ok, I've tried this and added your code. However, I get this from `mypy`:

```
    def __init__(self, version_string: str) -> Version:
NameError: name 'Version' is not defined
```

I guess, `__init__` is special here as it does NOT return any values. So I removed the `-> Version` part and tried it again. That was better, but I get this from `mypy`:

```
src/semver/version.py:122: error: Overloaded function implementation does not accept all possible arguments of signature 1
src/semver/version.py:122: error: Overloaded function implementation does not accept all possible arguments of signature 2
src/semver/version.py:382: error: No overload variant of "Version" matches argument type "int"
src/semver/version.py:382: note: Possible overload variants:
src/semver/version.py:382: note:     def __init__(self, version_string: str) -> Version
src/semver/version.py:382: note:     def __init__(self, major: SupportsInt, minor: SupportsInt, patch: SupportsInt) -> Version
src/semver/version.py:395: error: No overload variant of "Version" matches argument types "int", "int"
src/semver/version.py:395: note: Possible overload variants:
src/semver/version.py:395: note:     def __init__(self, version_string: str) -> Version
src/semver/version.py:395: note:     def __init__(self, major: SupportsInt, minor: SupportsInt, patch: SupportsInt) -> Version
src/semver/version.py:423: error: No overload variant of "Version" matches argument types "int", "int", "int", "str"
src/semver/version.py:423: note: Possible overload variants:
src/semver/version.py:423: note:     def __init__(self, version_string: str) -> Version
src/semver/version.py:423: note:     def __init__(self, major: SupportsInt, minor: SupportsInt, patch: SupportsInt) -> Version
src/semver/version.py:443: error: No overload variant of "Version" matches argument types "int", "int", "int", "Optional[str]", "str"
src/semver/version.py:443: note: Possible overload variants:
src/semver/version.py:443: note:     def __init__(self, version_string: str) -> Version
src/semver/version.py:443: note:     def __init__(self, major: SupportsInt, minor: SupportsInt, patch: SupportsInt) -> Version
src/semver/_deprecated.py:240: error: No overload variant of "Version" matches argument types "Any", "Any", "Any", "Any", "Any"
src/semver/_deprecated.py:240: note: Possible overload variants:
src/semver/_deprecated.py:240: note:     def __init__(self, version_string: str) -> Version
src/semver/_deprecated.py:240: note:     def __init__(self, major: SupportsInt, minor: SupportsInt, patch: SupportsInt) -> Version
Found 7 errors in 2 files (checked 7 source files)
```

It seems, it's difficult to use positional arguments and keyword arguments at the same time as https://stackoverflow.com/a/57225493 implies.

Any ideas?

src/semver/version.py
tomschr: You mean this should be ignored, right?

src/semver/version.py
tomschr: > I think we may have an architecture problem on our hands here. We shouldn't be calling `_ensure_str` on things other than a variation of string, so we ought to make sure we aren't passing in an `int`.

Yes, could be. That's why I didn't merge this PR yet. It is a bigger change and I would like to be fully sure and discuss that with you first. I wouldn't like to introduce any regression. :wink: 

Well, I thought, it could be an elegant way to have this line:

```python
prerelease = cls._ensure_str(prerelease or verlist[3])
```

That would either convert bytes -> str or left None and int untouched. Further down this case is handled:

```python
self._prerelease = None if prerelease is None else str(prerelease)
```

So here is the final assignment to `self._prerelease` to either store None or a string (regardless if it was passed as bytes, str, or int).

Maybe the name of the function is a bit irritating. Or we need to find a better way to assign `self._prerelease` its final value. Any ideas?

src/semver/version.py
tlaferriere: Nope that's the good one. The first review ( the one where I suggested to change the type hints of `_ensure_str`) is the one that's bad. So we should change it back to what it was and add this in to make the types correct by checking if it is an int.

tlaferriere: I'll put a sad face on the bad reviews

src/semver/version.py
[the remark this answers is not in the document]
tlaferriere: For the `Version` type hints, just put quotes around it. This is because it is not declared yet and the evaluation needs to be deferred when performing the type checks. I think in python 3.10 all type hints evaluation will be deferred by default, but for now we've got to use a string for that effect.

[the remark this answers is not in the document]
tlaferriere: For the overloads with the kwargs, I think adding more overloads for each possible combination of `prerelease` and `build` would help. Or maybe we can try something like `**kwargs: String` after the three positional parts.

src/semver/version.py
tomschr: Ahh, ok, thanks for the clarification. :+1: 

I'm a bit confused. Shouldn't prerelease and build only be None or str? At least our property say so:

```python
@property
def prerelease(self) -> Optional[str]:
```

So this line would probably raise errors in mypy:

```python
if isinstance(prerelease, int):
            self._prerelease = prerelease
```

Is my assumption correct?

src/semver/version.py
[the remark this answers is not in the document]
tomschr: > For the `Version` type hints, just put quotes around it.

I've tried that before. :wink: This is not possible as you get from `mypy`:

    error: The return type of "__init__" must be None

In other words: the `__init__` function does not have any return type at all (except None).

src/semver/version.py
tlaferriere: Well I want to consider also the fact that somebody might not want to provide the format but only the relevant information (the number). Although maybe we should keep that for later and simply consider the case of a string for now.

src/semver/version.py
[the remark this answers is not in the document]
tomschr: No idea how to solve that. :man_shrugging: Even if I add no return types to `__init__`I still get the above errors ("No overload variant of "Version" matches argument types...").

tomschr:
@tlaferriere Sorry for the delay, I was busy with other things.

I've played around a bit and introduced another function. I'm not completely happy, but the if clauses are kind of hidden inside a new `_enforce_str()` function. This makes the `__init__` function a bit more readable. Maybe the function could be better named (as they actually doesn't return a string, they can also return `None`, so... :thinking: ).

Apart from this, I'm not sure about the `@overload` decorator. Probably I'm doing something wrong or this complexity cannot be solved through this.

Although the `__init__` function gives more flexibility it also increased its complexity. So... :thinking:
On the one side, it would greatly enhance the usability. Developers don't have to use cumbersome methods, they just throw into the class whatever is appropriate. I like that.
On the other side, the initializer has 63 lines. That's a lot for such a function. Not sure if we can simplify that without compromising the usability.

Thoughts? :slightly_smiling_face:

PR #320, opened 2020-11-08
Fix #309: Make some functions private

### LINKED ISSUE #309: Rework some "private" functions in semver.version?
# Situation

Currently, we have the following module level functions inside `semver.version`:

* `comparator`: a decorator that wraps a version binary operator method to allow type checks. This is only used in the `Version`/`VersionInfo` class.

* `ensure_str`: a function to make sure we have always (Unicode) strings regardless if we pass a str or bytes type. Used in `semver.version.Version.parse` method.

* `cmd`: a function that is used inside `_nat_cmp` and `Version.compare`.

# Proposed Solution

IMHO the above function shouldn't be part of our public API as they are hardly used outside of semver. Their use cases are very limited. Additionally, we should discourage using these functions in user's code as they mainly act as helper functions for semver's `Version` class.

As such, I propose:

* Rename the above function and add a `_` (underscore) prefix. This marks this name  "private" and therefore it is not part of the public API anymore.

* Remove the function `ensure_str`. It was introduced just as a compatibility layer for Python2/3. IMHO it is better to integrate the code directly into `semver.version.Version.parse`.

* If possible and feasible, move `cmd` and `_nat_cmd` into the Version class.

As we are going to publish a new major release, I think this is a good opportunity to tackle this.

@python-semver/reviewers @tlaferriere what do you think? Any objections?

tomschr:
This PR fixes #309 and contains:

* Rename `comparator` -> `_comparator`, `cmp` -> `_cmp`
* Remove `ensure_str` and integrate it into `Version.parse`
* Rework tests in `test_typeerror-274.py`
* Move `_nat_cmp` to `Version` and make it a class method
* Remove private functions from API documentation

### DIFF
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -29,37 +29,7 @@ Comparable = Union["Version", Dict[str, VersionPart], Collection[VersionPart], s
 Comparator = Callable[["Version", Comparable], bool]
 
 
-def cmp(a, b):  # TODO: type hints
-    """Return negative if a<b, zero if a==b, positive if a>b."""
-    return (a > b) - (a < b)
-
-
-def ensure_str(s: String, encoding="utf-8", errors="strict") -> str:
-    # Taken from six project
-    """
-    Coerce *s* to `str`.
-
-    * `str` -> `str`
-    * `bytes` -> decoded to `str`
-
-    :param s: the string to convert
-    :param encoding: the encoding to apply, defaults to "utf-8"
-    :param errors: set a different error handling scheme,
-       defaults to "strict".
-       Other possible values are `ignore`, `replace`, and
-       `xmlcharrefreplace` as well as any other name
-       registered with :func:`codecs.register_error`.
-    :raises TypeError: if ``s`` is not str or bytes type
-    :return: the converted string
-    """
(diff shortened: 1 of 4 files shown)


src/semver/version.py
```diff
@@ -78,31 +48,9 @@ def wrapper(self: "Version", other: Comparable) -> bool:
...
-
-    a, b = a or "", b or ""
-    a_parts, b_parts = split_key(a), split_key(b)
-    for sub_a, sub_b in zip(a_parts, b_parts):
-        cmp_result = cmp_prerelease_tag(sub_a, sub_b)
-        if cmp_result != 0:
-            return cmp_result
-    else:
-        return cmp(len(a), len(b))
+def _cmd(a, b):  # TODO: type hints
```
tlaferriere, line 51: I believe you meant `_cmp` 😉

tlaferriere (changes requested):

src/semver/version.py
tomschr: Oh yes, of course! Will correct it. Thanks for spotting this stupid error. :+1:

PR #337, opened 2021-04-21
Improve suggestion for finding max and min using semver

mzjp2:
I was reading through the documentation and thought that the suggestion for finding max/min of a list/iterable of strings using `semver` unecessarily complicated, with a `map`, and multiple typecasts. I think using just `max` with `key` as done in this PR is much cleaner and should be the de-facto suggestion.

This is my first contribution to this repository, I hope it's welcome and I've followed the contribution guide well - but feel free to let me know if I can improve anything or close this PR if it's unwanted :)

### DIFF
--- a/src/semver/_types.py
+++ b/src/semver/_types.py
@@ -1,5 +1,6 @@
 """Typing for semver."""
 
+from functools import partial
 from typing import Union, Optional, Tuple, Dict, Iterable, Callable, TypeVar
 
 VersionPart = Union[int, Optional[str]]
@@ -8,3 +9,4 @@ VersionDict = Dict[str, VersionPart]
 VersionIterator = Iterable[VersionPart]
 String = Union[str, bytes]
 F = TypeVar("F", bound=Callable)
+Decorator = Union[Callable[..., F], partial]
--- a/docs/migratetosemver3.rst
+++ b/docs/migratetosemver3.rst
@@ -20,7 +20,7 @@ Use Version instead of VersionInfo
 The :class:`VersionInfo` has been renamed to :class:`Version`
 to have a more succinct name.
 An alias has been created to preserve compatibility but
-using old name has been deprecated.
+using the old name has been deprecated.
 
 If you still need the old version, use this line:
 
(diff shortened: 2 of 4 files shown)


tomschr:
Thank you very much @mzjp2 and welcome! Much appreciated. :+1:
You are right, your idea is much easier. I like it! 

For some reason, there is an (unrelated) issue about the typing module. I'm not sure about the problem. @tlaferriere as you are more experienced than I am, could you have a look, please?

After the problem is fixed, I'd like to merge it.

If you have further ideas, don't hesitate, just submit your suggestions. Thank you!

mzjp2:
Thank you for the fix!

tomschr:
@mzjp2 @tlaferriere I've played a bit with the type definitions. For some reason, mypy seems to detect an error. I remember, this was different in the past, so they changed something. To avoid such thing in the future, maybe it would be useful to freeze the version.

I've introduced a new type `DecoratorType` and with that it seems to work. Maybe there is a better name.

Feel free to suggest a better name or handling.

mzjp2:
Sounds good to me - happy for you to merge whenever you want :)

tomschr:
Let's wait some more day, so Thomas has some time to look into it. :slightly_smiling_face:

tlaferriere:
I'll look into this tomorrow, today I have an exam.

tomschr:
Good luck with the exam! Keep my fingers crossed. :crossed_fingers:

mzjp2:
Good luck indeed!

>

src/semver/_types.py
```diff
@@ -8,3 +9,4 @@
 VersionIterator = Iterable[VersionPart]
 String = Union[str, bytes]
 F = TypeVar("F", bound=Callable)
+DecoratorType = Union[Callable[..., F], partial]
```
tlaferriere, line 12: @tomschr Good idea to add this alias, we just don't need the `Type` suffix since it is clear it is a type from it's import path.
```suggestion
Decorator = Union[Callable[..., F], partial]
```

tlaferriere (commented):
Well spotted @mzjp2 this is much cleaner than before.
@tomschr I haven't seen any mypy failure, I assume you fixed it? Anyway, apart from the suggestion, I approve this PR.

tomschr:
Thanks Thomas and Zain for your contribution, much appreciated! :+1: :heart:

mzjp2:
Thanks all!

PR #343, opened 2021-11-06
Added pydantic integration for Version

calebstewart:
This PR adds native [pydantic](https://pydantic-docs.helpmanual.io/) integration to `python-semver`. Pydantic makes defining data schemas/models super easy and intuitive, but doesn't have a native field type for semantic version strings. Custom field types are created by simply including a class method `__get_validators__` on the class which yields methods to validate input syntax and return the resulting type. In this case, the `parse` method functions naturally as the field validator. Pydantic also allows the `__modify_schema__` method to inject extra details about the field into the model schema.

In this case, I simply added the `__modify_schema__` and `__get_validators__` methods to the `Version` class. These two class-methods are one line each. The former injects semantic versioning examples into the generated schema while the latter yields the `Version.parse` method as a validator. The result is that a Pydantic user can do the following:

```py
import pydantic
import semver

class MyModel(pydantic.BaseModel):
  version: semver.Version

# model.version will be an instance of semver.Version
# and the construction will raise a pydantic.ValidationError
# if improperly formatted.
model = MyModel.parse_obj({"version": "1.2.3"})
```

It's worth noting that a `pydantic` user could manually specify `Version.parse` as a validator for the field in their model, but this is less expressive, and the overhead for making `semver.Version` compatible with `pydantic` is relatively low (two one-line class methods).

That being said, I totally understand if you think this is out of scope here, but it seemed like a simple quality of life improvement, and thought I'd throw it out there. I went through the contributing guidelines, and I think I've hit all the marks. Thanks for reviewing/considering! :smiley:

Changes:
- Added two new dunder methods in Version which comply with the custom
field types described [here](https://pydantic-docs.helpmanual.io/usage/types/#classes-with-__get_validators__).
- Updated usage documentation with example for using with `pydantic`.
- Added test case for `pydantic` integration (mainly copied parsing tests, and modified to test `pydantic` integration).
- Added `pydantic` dependency in `tox.ini` for tests cases.
- Verified all tests passed for all python versions with `tox`.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -787,3 +787,18 @@ the original class:
      Traceback (most recent call last):
      ...
      ValueError: '1.2.4': not a valid semantic version tag. Must start with 'v' or 'V'
+
+Integrating with Pydantic
+-------------------------------------
+
+If you are building a Pydantic model, you can use :class:`~semver.version.Version` directly.
+An appropriate :class:`pydantic.ValidationError` will be raised for invalid version numbers.
+
+.. code-block:: python
+
+    >>> from semver.version import Version
+    >>> from pydantic import create_model
+    >>> Model = create_model('Model', version=(Version, ...))
+    >>> model = Model(version="3.4.5-pre.2+build.4")
+    >>> model.version
+    Version(major=3, minor=4, patch=5, prerelease='pre.2', build='build.4')
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -1,28 +1,22 @@
 """Version handling."""
 
-import collections
 import re
-from functools import wraps
+import collections
 from typing import (
     Any,
     Dict,
-    Iterable,
-    Optional,
-    SupportsInt,
     Tuple,
     Union,
-    cast,
     Callable,
+    Iterable,
+    Optional,
     Collection,
+    SupportsInt,
+    cast,
 )
+from functools import wraps
 
-from ._types import (
-    VersionTuple,
-    VersionDict,
-    VersionIterator,
-    String,
-    VersionPart,
-)
+from ._types import String, VersionDict, VersionPart, VersionTuple, VersionIterator
 
 # These types are required here because of circular imports
 Comparable = Union["Version", Dict[str, VersionPart], Collection[VersionPart], str]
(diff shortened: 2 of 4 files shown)


tomschr (changes requested):
Thank you @calebstewart for your efforts and sorry for the delay! Much appreciated.

I never heard about Pydantic, but it seems to be an interesting project. Not entirely sure if this is something for this project, but I think we should give it a try. :slightly_smiling_face: 

I've added some small, minor comments. If you could have a look we could merge them.

tomschr:
After looking at this PR again and thinking about it, I'm sorry but I'm not entirely convinced we should amend the class with these methods.

Although the overall change is simple and doesn't introduce any runtime dependencies (which is good!), I have the feeling it doesn't add any value to non-Pydantic users. How many users really use semver and pydantic in this way? What should we do if another user have the same idea, but with a different library? Or what happens if Pydantic introduces some incompatible changes? Where do we draw the line? Wouldn't that lead to a "bloated" class?

I know, it's a bit of a stretch. However, my philosophy in this regard is that I try to keep any runtime dependencies away from semver. It shouldn't depend on other libraries as the overall scope is quite focused. So in other words, it should stand on its own legs. Maybe we should follow this approach with "virtual" dependencies too.

Have you tried to solve your issue by inheritance?

```python
class PydanticVersion(Version):
   @classmethod
    def __get_validators__(cls):
        """Return a list of validator methods for pydantic models."""
        yield cls.pars

    @classmethod
    def __modify_schema__(cls, field_schema):
        """Inject/mutate the pydantic field schema in-place."""
        field_schema.update(examples=["1.0.2", "2.15.3-alpha", "21.3.15-beta+12345"])
```

Then you could add this to your model:

```python
class MyModel(pydantic.BaseModel):
  version: PydanticVersion
```

I haven't tried it, but it looks like a small cost.

However, I don't want to completely reject it. I'd like to suggest to include it into the documentation. It could serve as an example about how we could combine semver with other projects.

Let me know if you have a compelling argument that I've overlooked. I'd really like to know! I leave this PR open for the time being.

calebstewart:
I apologize for never getting back to you on those requests. The holidays were kind of crazy, and this fell off my plate. I totally understand the apprehension here. I wasn't completely sure if it fit either, to be honest. Not a huge deal.

I think your example looks right as well. That is what I ended up doing in the end. If you'd like to add something like that as a note/example in the documentation, I think that would be really cool and possibly helpful! I have no problem with this PR being closed, though. I totally understand your position. :+1:

tomschr:
All good, don't worry. I was also busy with other stuff. :wink: 

Thank you Caleb for your nice words! I'm really sorry, hope it wasn't too much effort on your side.

I'd proceed as follows: I'll integrate a new section into the documentation describing the above finding. @calebstewart Would you like to review it once it's finished?

tomschr:
Closing this PR in favor of #353

PR #353, opened 2022-01-27
Describe Pydantic and semver in "Advanced topics"

tomschr:
This PR adds a new section into "Advanced topic". Fixing #343.

@calebstewart would you mind having a look? Anything that is missing, wrong, or could be done better? Would be really helpful. Thank you very much! :heart:

### DIFF
--- a/changelog.d/343.doc.rst
+++ b/changelog.d/343.doc.rst
@@ -1,0 +1,2 @@
+Describe combining Pydantic with semver in the "Advanced topic"
+section.
--- a/docs/advanced/combine-pydantic-and-semver.rst
+++ b/docs/advanced/combine-pydantic-and-semver.rst
@@ -1,0 +1,53 @@
+Combining Pydantic and semver
+=============================
+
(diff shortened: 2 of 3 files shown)


tomschr:
If you don't object, I will merge it to master tomorrow. :slightly_smiling_face:

calebstewart:
I just double checked the code. My only note is that the `__get_validators__` method should be `yield cls.parse` instead of `yield cls.pars`. Just a missing `e`.

Aside from that, the examples work, and make sense to me! Thanks for checking in with me! I appreciate it!

tomschr:
> I just double checked the code. My only note is that the `__get_validators__` method should be `yield cls.parse` instead of `yield cls.pars`. Just a missing `e`.

Ahh, good catch! :+1:  Thanks for the hint, I've already fixed it.

> Aside from that, the examples work, and make sense to me! Thanks for checking in with me! I appreciate it!

Thank you! I've added you as a co-author. :slightly_smiling_face:

tomschr:
The documentation is now available at https://python-semver.readthedocs.io/en/latest/advanced/combine-pydantic-and-semver.html

PR #355, opened 2022-02-24
Use HTTPS instead of HTTP for the website URL

b0uh:
Hello,

I noticed that all the URLs to be website were using HTTP instead of HTTPS. So I opened this PR to fix that since we are redirected to the HTTPS URL anyway.

If that was intended to use HTTP, please ignore this PR.


tomschr:
Thank you @b0uh! Good catch! :+1: Of course it would be better to use HTTPS instead of HTTP.

b0uh:
Just notice HTTP is also used in the "About" section of this repo (:
Screenshot 2022-02-24 at 14 22 56

tomschr:
Ahh, thanks! That tiny little thing. Easy to be overlooked. Ok, fixed that too.
Thank you! :heart:

PR #359, opened 2022-05-13
Allow shorter version for parsing

OidaTiftla:
For my personal use I added the possibility to parse a version string that is shorter than `major.minor.patch`, for example `13.5`.

When creating a semver version object using the constructor it is possible to only pass a major version and the other parameters are filled with a default value. So why not give this possibility to the parsing function.

To keep backwards compatibility and the requirement from semver that normal version strings need at least three version numbers, the parsing of shorter version strings must be enabled with a parameter. Without that parameter the old functionality will be executed.

I do not know if this is something you would like to integrate, if not simply close this pull request. Otherwise give me a short feedback, then I will update the documentation and the test suite.

### DIFF
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -68,29 +68,8 @@ class Version:
     __slots__ = ("_major", "_minor", "_patch", "_prerelease", "_build")
     #: Regex for number in a prerelease
     _LAST_NUMBER = re.compile(r"(?:[^\d]*(\d+)[^\d]*)+")
-    #: Regex for a semver version
-    _REGEX = re.compile(
-        r"""
-            ^
-            (?P<major>0|[1-9]\d*)
-            \.
-            (?P<minor>0|[1-9]\d*)
-            \.
-            (?P<patch>0|[1-9]\d*)
-            (?:-(?P<prerelease>
-                (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
-                (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
-            ))?
-            (?:\+(?P<build>
-                [0-9a-zA-Z-]+
-                (?:\.[0-9a-zA-Z-]+)*
-            ))?
-            $
-        """,
-        re.VERBOSE,
-    )
-    #: Regex for a semver version that might be shorter
-    _REGEX_ALLOW_SHORT = re.compile(
+    #: Regex template for a semver version
+    _REGEX_TEMPLATE = \
         r"""
             ^
             (?P<major>0|[1-9]\d*)
@@ -100,8 +79,8 @@ class Version:
                 (?:
                     \.
                     (?P<patch>0|[1-9]\d*)
-                )?
-            )?
+                ){}
+            ){}
             (?:-(?P<prerelease>
                 (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
                 (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
@@ -111,7 +90,15 @@ class Version:
                 (?:\.[0-9a-zA-Z-]+)*
             ))?
             $
-        """,
+        """
+    #: Regex for a semver version
+    _REGEX = re.compile(
+        _REGEX_TEMPLATE.format('', ''),
+        re.VERBOSE,
+    )
+    #: Regex for a semver version that might be shorter
+    _REGEX_OPTIONAL_MINOR_AND_PATCH = re.compile(
+        _REGEX_TEMPLATE.format('?', '?'),
         re.VERBOSE,
     )
 
@@ -578,7 +565,7 @@ build='build.10')
         return cmp_res in possibilities
 
     @classmethod
-    def parse(cls, version: String, allowShorterVersion: bool = False) -> "Version":
+    def parse(cls, version: String, optional_minor_and_patch: bool = False) -> "Version":
         """
         Parse version string to a Version instance.
 
@@ -587,6 +574,7 @@ build='build.10')
            allow subclasses.
 
         :param version: version string
+        :param optional_minor_and_patch: if set to true, the minor and patch version digits are optional, but this will deviate from the semver spec
         :return: a new :class:`Version` instance
         :raises ValueError: if version is invalid
         :raises TypeError: if version contains the wrong type
@@ -600,8 +588,8 @@ prerelease='pre.2', build='build.4')
         elif not isinstance(version, String.__args__):  # type: ignore
             raise TypeError("not expecting type '%s'" % type(version))
 
-        if allowShorterVersion:
-            match = cls._REGEX_ALLOW_SHORT.match(version)
+        if optional_minor_and_patch:
+            match = cls._REGEX_OPTIONAL_MINOR_AND_PATCH.match(version)
         else:
             match = cls._REGEX.match(version)
         if match is None:
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -79,8 +79,8 @@ class Version:
                 (?:
                     \.
                     (?P<patch>0|[1-9]\d*)
-                ){}
-            ){}
+                ){opt_patch}
+            ){opt_minor}
             (?:-(?P<prerelease>
                 (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
                 (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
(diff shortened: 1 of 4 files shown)


tomschr:
Thanks @OidaTiftla for your contribution! And sorry for the delay.

It looks good. My only concern is the repetition of the content between the class variables `_ REGEX` and `_REGEX_ALLOW_SHORT `. Both share a lot of similar strings. As far as I can see, the only distinction is that the patch part is optional, right?

If feasible, could be combine these class variables somehow?

tomschr:
I looked a bit closer and I see the following issues:

1. As mentioned before, the two class variables violate the DRY principle.
1. The variable name `allowShorterVersion` is camel case style. This is Python, we should avoid that. :wink:
1. The docstring misses an explanation of the second variable variable of the `parse()` method.

My idea to consolidate the two class variables would be to define a placeholder for the patch part like this:

```python
_REGEX = (
    r"""
        ^
        (?P<major>0|[1-9]\d*)
        (?:
            \.
            (?P<minor>0|[1-9]\d*)
            (?:
                \.
                (?P<patch>0|[1-9]\d*)
            ){}
        )
        (?:-(?P<prerelease>
            (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
            (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
        ))?
        (?:\+(?P<build>
            [0-9a-zA-Z-]+
            (?:\.[0-9a-zA-Z-]+)*
        ))?
        $
    """
)
```

In the `parse()` method, we need to decide, if we want the optional patch part or not. I have something like this in my mind:


```python
def parse((cls, version: String, optional_patch: bool = False) -> "Version":
    regex = re.compile(_REGEX.format("?" if optional_patch else ""),  re.VERBOSE)
    match = regex.match(version)
    if match is None:
         raise ValueError(f"{version} is not valid SemVer string")

     matched_version_parts: Dict[str, Any] = match.groupdict()
     if not matched_version_parts['minor']:
          matched_version_parts['minor'] = 0
     if not matched_version_parts['patch']:
          matched_version_parts['patch'] = 0

     return cls(**matched_version_parts)
)

    match = regex.match(version)
```

This is just a rough draft. I'm not sure if this would be useful as we will loose the `re.compile` optimization.

src/semver/version.py
```diff
@@ -68,15 +68,19 @@ class Version:
...
-            \.
-            (?P<patch>0|[1-9]\d*)
+            (?:
+                \.
+                (?P<minor>0|[1-9]\d*)
+                (?:
+                    \.
+                    (?P<patch>0|[1-9]\d*)
+                ){}
+            ){}
```
tomschr, line 83: To make it a bit more explicit, I would introduce some placeholders:

```suggestion
                ){opt_patch}
            ){opt_minor}
```

src/semver/version.py
```diff
@@ -86,7 +90,15 @@ class Version:
...
-        """,
+        """
+    #: Regex for a semver version
+    _REGEX = re.compile(
+        _REGEX_TEMPLATE.format('', ''),
+        re.VERBOSE,
+    )
+    #: Regex for a semver version that might be shorter
+    _REGEX_OPTIONAL_MINOR_AND_PATCH = re.compile(
+        _REGEX_TEMPLATE.format('?', '?'),
```
tomschr, line 101: As we've introduced these placeholders before, we can now use them:

```suggestion
        _REGEX_TEMPLATE.format(opt_patch='', opt_minor=''),
        re.VERBOSE,
    )
    #: Regex for a semver version that might be shorter
    _REGEX_OPTIONAL_MINOR_AND_PATCH = re.compile(
        _REGEX_TEMPLATE.format(opt_patch='?', opt_minor='?'),
```

tomschr:
@OidaTiftla Thanks for your contribution! :+1: 

>  >  Thanks @OidaTiftla for your contribution! And sorry for the delay.
>
> Everything okay. It's open source, so don't stress.

Haha, yes, you're right.


> > 1.  As mentioned before, the two class variables violate the DRY principle.
>
> I fully support this +1, the format string solution is a very nice idea.

Thanks. I would even go further and be _explicit_. In my example, I wasn't and that could be improved with a named placeholder.


> > Apart from the technical issue, there is a fundamental question. Should we deviate from the semver spec? Should we allow parsing of incomplete version strings? I'm not sure as we already have this documented here: https://python-semver.readthedocs.io/en/latest/advanced/deal-with-invalid-versions.html
> > Have you seen this?
> 
> I was reading though the docs but didn't found that part before trying to adjust the code to fit my needs. So thanks for the link. 

You're welcome. :slightly_smiling_face: 

> What does not make me happy about this workaround is, that I have to do the parsing by my own and cannot use the library for that part. So I would copy the regex from the source code of python_semver.

Not sure if I understand it correctly, but why copying the regex? Wouldn't that be something for creating a subclass from `Version`? In the derived class you can access the regex from the base class. 


> My opinion to the fundamental question: If the method is used with the default parameters it will still throw an error if it is not a valid semver. So I think it would be okay, but that's something you or you with the other maintainers must decide if you want that optional parameter to deviate from the spec. Users would not need the workaround from above and can still benefit from that library. For me it would be totally fine, if you close this PR and I use my fork in my current project.

Thanks for your thoughts. :+1: 

I think, I would be okay if a) the default behavior isn't changed, b) it's clearly documented in the docstring, and c) all tests are green. You already did a and b. Look at c, there are some overly lone lines, but nothing serious.

What I would like to avoid is to blow up the library to an "all-purpose-parse-whatever-shitty-version-string-you have". :wink: There is 


> > This is just a rough draft. I'm not sure if this would be useful as we will loose the `re.compile` optimization.
> 
> I've fixed the DRY violation a bit differently, you may have a look at it and report if this would keep the `re.compile` optimisation.

I will come up with an additional review.

> If you do not want to deviate from the semver spec it's also ok for me, you can simply close this PR. I will then stick to my fork for my current project, that's also fine.

I'm happily integrate your change. Just a few minor things that I will come up in a few seconds.

tomschr (changes requested):
@OidaTiftla Just some minor things I would like to see. If you agree, I can integrate it.

Many thanks!

OidaTiftla:
Thanks for your feedback @tomschr! 👍

> > What does not make me happy about this workaround is, that I have to do the parsing by my own and cannot use the library for that part. So I would copy the regex from the source code of python_semver.
> 
> Not sure if I understand it correctly, but why copying the regex? Wouldn't that be something for creating a subclass from `Version`? In the derived class you can access the regex from the base class.

Yes you're right, a derived class would have access to it, but I was not explicit enough, sorry for that. My statement of copy the regex will not solve the problem, to parse version strings with optional patch and minor version parts. One must copy the regex from python_semver and adjust it. Hope I could clarify my thought a bit.

> I think, I would be okay if a) the default behavior isn't changed, b) it's clearly documented in the docstring, and c) all tests are green. You already did a and b. Look at c, there are some overly lone lines, but nothing serious.

Ok. I'll look into it.

> What I would like to avoid is to blow up the library to an "all-purpose-parse-whatever-shitty-version-string-you have". 😉 There is

Yes, I'm totally on your site. And you as maintainer should always ask for the necessity of changes and keep the library clean 👍

src/semver/version.py
OidaTiftla: Yes, great idea 👍

src/semver/version.py
```diff
@@ -562,6 +574,7 @@ def parse(cls, version: String) -> "Version":
            allow subclasses.
 
         :param version: version string
+        :param optional_minor_and_patch: if set to true, the minor and patch version digits are optional, but this will deviate from the semver spec
```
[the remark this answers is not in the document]
OidaTiftla: Thanks, I did not know how to make a line break, so that the doc generator will read it correctly. And it is much clearer to understand now.

OidaTiftla:
@tomschr can you please trigger the workflows again? So I can see if I could fix everything. Thanks in advance.

Locally in my environment there were not more errors when running `tox --skip-missing-interpreters`

And are the added tests enough or should there be more unit tests added?

tomschr:
> > Not sure if I understand it correctly, but why copying the regex? Wouldn't that be something for creating a subclass from `Version`? In the derived class you can access the regex from the base class.
> 
> Yes you're right, a derived class would have access to it, but I was not explicit enough, sorry for that. My statement of copy the regex will not solve the problem, to parse version strings with optional patch and minor ve

Ahh, right! Thanks for the clarification!


> can you please trigger the workflows again? So I can see if I could fix everything. Thanks in advance.

Done. All green! :wink: 

> Locally in my environment there were not more errors when running `tox --skip-missing-interpreters`

That's fine. Just in case: you can use [`pyenv`](https://github.com/pyenv/pyenv) if you need different versions of the Python interpreters. Quite handy!

> And are the added tests enough or should there be more unit tests added?

They are fine. :+1: 

I realized, I haven't looked at some parts in the documentation/changelog. I would like to add some small additions. If that's done, I'll merge them. Promised! :slightly_smiling_face:

tomschr (changes requested):
@OidaTiftla Just some minor additions. Would be great if you could include them. Thanks!

OidaTiftla:
@tomschr thanks for your feedback.

> > Locally in my environment there were not more errors when running `tox --skip-missing-interpreters`
> 
> That's fine. Just in case: you can use [`pyenv`](https://github.com/pyenv/pyenv) if you need different versions of the Python interpreters. Quite handy!

Ah thank you for this, this would help 👍 

> I realized, I haven't looked at some parts in the documentation/changelog. I would like to add some small additions. If that's done, I'll merge them. Promised! 🙂

It was my fault to read the `CONTRIBUTION.rst` that late. I happily integrate your suggestions, that's no problem. You did the work, thanks for your help 😄

tomschr:
@OidaTiftla Many thanks for your contributions! :+1:

If you have further ideas or suggestions I would happily discuss it with you. :slightly_smiling_face:  There are also some open issues that would need some love and dedication. Also I'm working on getting out a 3.0.0 release. Not sure when this will be ready.

PR #362, opened 2022-07-03
Support matching 'equal' when no operator is provided

jmi2k:
This makes it easier to work with user-supplied versions, as they may omit the comparison operator and just specify which version they want to select. Equality comparison is a sane default behavior.

### DIFF
--- a/changelog.d/pr362.feature.rst
+++ b/changelog.d/pr362.feature.rst
@@ -1,0 +1,2 @@
+Make :meth:`.Version.match` accept a bare version string as match expression, defaulting to
+equality testing.
--- a/docs/usage/compare-versions-through-expression.rst
+++ b/docs/usage/compare-versions-through-expression.rst
@@ -24,3 +24,16 @@ That gives you the following possibilities to express your condition:
     True
     >>> semver.match("1.0.0", ">1.0.0")
     False
+
+If no operator is specified, the match expression is interpreted as a
+version to be compared for equality. This allows handling the common
+case of version compatibility checking through either an exact version
+or a match expression very easy to implement, as the same code will
+handle both cases:
+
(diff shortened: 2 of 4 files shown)


tomschr:
Hi @jmi2k! Thanks for your contribution, very much appreciated. :+1: :rocket: 

Your implementation looks good, I only have two wishes to complete it:

1. **Amend the documentation.**
   The documentation should explain this new functionality too. You could add it to the section [_Comparing Versions through an Expression_](https://python-semver.readthedocs.io/en/latest/usage/compare-versions-through-expression.html). Find the source code in `docs/usage/compare-versions-through-expression.rst`.
   Could you amend the doc and add a short example? Bonus points if you could give or explain the differences to `Version.parse("a.b.c") == Version.parse("d.e.f")`. What approach do we want to recommend to our developers? (The Zen of Python, states, "There should be one-- and preferably only one --obvious way to do it.")
2. **Create a changelog entry**
   Add a short snippet which explains your addition (also mentioning this pull request number). You can find the details how to create such an entry in the section [Adding a Changelog Entry](https://python-semver.readthedocs.io/en/latest/development.html#adding-a-changelog-entry).

Many thanks!

jmi2k:
First of all, thanks for the in-depth response! I highly appreciate it, as I was kinda lost about how to add that changelog entry.

Now I think I have everything in place. The rationale behind this change is that I'm building a plugin dependency system for some in-house tool, and as you will know for sure it's pretty common to specify just the version as an exact dependency. As for now, I've written code around the access to patch the equality operator in when no operator is provided, but I felt like that it's more elegant for the library to handle that case too (as the changes required were minimal).

If you think I should change anything else, please don't hesitate to tell me!

PR #363, opened 2022-07-10
Updated docs with VersionInfo.

max-pfeiffer:
Hi guys, I just made a little update to the doc section in README.rst. Might save somebody some time digging through the source to find out about that change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,11 +47,11 @@ To import this library, use:
     >>> import semver
 
 Working with the library is quite straightforward. To turn a version string into the
-different parts, use the ``semver.Version.parse`` function:
+different parts, use the ``semver.VersionInfo.parse`` function:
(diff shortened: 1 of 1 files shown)


tomschr:
Thank you @max-pfeiffer for your contribution. Much appreciated! :+1: 

However, `VersionInfo` was renamed to `Version` for semver3. It's still available (actually, it's an alias), but is considered deprecated. In future versions it might be removed.

This part is already documented in the chapter _Migrating from semver2 to semver3_ under the [Use Version instead of VersionInfo]( https://python-semver.readthedocs.io/en/latest/migration/migratetosemver3.html#use-version-instead-of-versioninfo).

As such, we shouldn't recommend the wrong name to our users :wink:  so closing it.

PR #367, opened 2022-11-15
Fix #284: Implement tilde and caret matching

### LINKED ISSUE #284: Feature request: concise "compatibility" matching
I'd really like a concise expression for "is version x compatible with version y?"

Specifically, suppose I write code against one version of an API, and the API exposes a semver-compliant version. Is `v_caller` compatible with `v_api`?

The most concise expression I can come up with is something like `v_caller <= v_api and v_caller.major == v_api.major`, but I really wish this was simpler and more readable. Maybe `v_caller.compatible(v_api)`; or maybe `v_caller.can_call(v_api)` better illustrates which operand is which.

I know this would be covered by #241, but its scope is quite broad. I wanted to specifically call out this feature.

tomschr:
* Introduce `Spec` class to deal with such comparisons
* Improve documentation
* Simplify code in Version.match (delegates to `Spec.match`)


Lexicality:
My specific use-case was an event processor that routed messages based on their semver number, so for example, everything matching `~=1.2` would go to the latest event handler, `==1.1.*` to a legacy handler and so on.
Ultimately I hacked something together and have since left the company, so it doesn't really matter to me anymore 😅

However, I will say that I really don't like operator overloading and much prefer a `match` method or similar.

tomschr:
> However, I will say that I really don't like operator overloading and much prefer a `match` method or similar.

Yeah, that's also one of the questions where I'm unsure about that. :thinking:

PR #368, opened 2022-11-16
Fix #284: implement "is compatible with" method

### LINKED ISSUE #284: Feature request: concise "compatibility" matching

tomschr:
This is a first implementation of `Version.is_compatible`.

The algorithm checks:

* if the two majors are different, it's incompatible
* if the two majors are equal, but the minor of the callee is lower than the caller, it's incompatible
* if both prereleases are present and different, it's incompatible
* otherwise it's compatible

@rafalkrupinski /@sbrudenell / @Lexicality maybe you want to try/review it? :slightly_smiling_face: 

# TODOs
* [x] Improve changelog
* [x] Improve docstring with example(s)
* [x] Add missing documentation
* [x] Review code
* [x] Clarify naming

### DIFF
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -647,7 +647,7 @@ prerelease='pre.2', build='build.4')
             raise TypeError(error)
 
     @classmethod
-    def isvalid(cls, version: str) -> bool:
+    def is_valid(cls, version: str) -> bool:
         """
         Check if the string is a valid semver version.
 
@@ -662,6 +662,42 @@ prerelease='pre.2', build='build.4')
             return True
         except ValueError:
             return False
+
+    def is_compatible(self, other: "Version") -> bool:
+        """
+        Check if current version is compatible with other version.
+
+        The result is True, if either of the following is true:
+
+        * both versions are equal, or
+        * both majors are equal and higher than 0. Same for both minors.
+          Both pre-releases are equal, or
+        * both majors are equal and higher than 0. The minor of b's
+          minor version is higher then a's. Both pre-releases are equal.
+
+        The algorithm does *not* check patches.
+
+        :param other: the version to check for compatibility
+        :return: True, if ``other`` is compatible with the old version,
+                 otherwise False
+
+        >>> Version(1, 1, 0).is_compatible(Version(1, 0, 0))
+        False
+        >>> Version(1, 0, 0).is_compatible(Version(1, 1, 0))
+        True
+        """
+        if not isinstance(other, Version):
+            raise TypeError(f"Expected a Version type but got {type(other)}")
+
+        # All major-0 versions should be incompatible with anything but itself
+        if (0 == self.major == other.major) and (self[:4] != other[:4]):
+            return False
+
+        return (
+            (self.major == other.major)
+            and (other.minor >= self.minor)
+            and (self.prerelease == other.prerelease)
+        )
 
 
 #: Keep the VersionInfo name for compatibility
--- a/tests/test_semver.py
+++ b/tests/test_semver.py
@@ -73,10 +73,65 @@ def test_should_be_able_to_use_integers_as_prerelease_build():
 
 
 def test_should_versioninfo_isvalid():
-    assert Version.isvalid("1.0.0") is True
-    assert Version.isvalid("foo") is False
+    assert Version.is_valid("1.0.0") is True
+    assert Version.is_valid("foo") is False
 
 
 def test_versioninfo_compare_should_raise_when_passed_invalid_value():
     with pytest.raises(TypeError):
         Version(1, 2, 3).compare(4)
+
+
+@pytest.mark.parametrize(
+    "old, new",
+    [
+        ((1, 2, 3), (1, 2, 3)),
+        ((1, 2, 3), (1, 2, 4)),
+        ((1, 2, 4), (1, 2, 3)),
+        ((1, 2, 3, "rc.0"), (1, 2, 4, "rc.0")),
+        ((0, 1, 0), (0, 1, 0)),
+    ],
+)
+def test_should_succeed_compatible_match(old, new):
+    old = Version(*old)
+    new = Version(*new)
+    assert old.is_compatible(new)
+
+
+@pytest.mark.parametrize(
+    "old, new",
+    [
+        ((1, 1, 0), (1, 0, 0)),
+        ((2, 0, 0), (1, 5, 0)),
+        ((1, 2, 3, "rc.1"), (1, 2, 3, "rc.0")),
+        ((1, 2, 3, "rc.1"), (1, 2, 4, "rc.0")),
+        ((0, 1, 0), (0, 1, 1)),
+        ((1, 0, 0), (1, 0, 0, "rc1")),
+        ((1, 0, 0, "rc1"), (1, 0, 0)),
+    ],
+)
+def test_should_fail_compatible_match(old, new):
+    old = Version(*old)
+    new = Version(*new)
+    assert not old.is_compatible(new)
+
+
+@pytest.mark.parametrize(
+    "wrongtype",
+    [
+        "wrongtype",
+        dict(a=2),
+        list(),
+    ],
+)
+def test_should_fail_with_incompatible_type_for_compatible_match(wrongtype):
+    with pytest.raises(TypeError, match="Expected a Version type .*"):
+        v = Version(1, 2, 3)
+        v.is_compatible(wrongtype)
+
+
+def test_should_succeed_with_compatible_subclass_for_is_compatible():
+    class CustomVersion(Version):
+        ...
+
+    assert CustomVersion(1, 0, 0).is_compatible(Version(1, 0, 0))
--- a/.gitignore
+++ b/.gitignore
@@ -268,3 +268,6 @@ fabric.properties
 
 docs/_api
 !docs/_api/semver.__about__.rst
+
+# For node
+node_modules/
--- a/changelog.d/284.deprecation.rst
+++ b/changelog.d/284.deprecation.rst
@@ -1,0 +1,5 @@
+Deprecate the use of :meth:`Version.isvalid`.
+
+Rename :meth:`Version.isvalid <semver.version.Version.isvalid>`
+to :meth:`Version.is_valid <semver.version.Version.is_valid>`
+for consistency reasons with :meth:`Version.is_compatible <semver.version.Version.is_compatible>`
--- a/changelog.d/284.doc.rst
+++ b/changelog.d/284.doc.rst
@@ -1,0 +1,1 @@
+Document deprecation of :meth:`Version.isvalid`.
--- a/changelog.d/284.feature.rst
+++ b/changelog.d/284.feature.rst
@@ -1,0 +1,1 @@
+Implement :meth:`Version.is_compatible <semver.version.Version.is_compatible>` to make "is self compatible with X".
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -118,8 +118,8 @@ todo_include_todos = False
 # Markup to shorten external links
 # See https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html
 extlinks = {
-    "gh": ("https://github.com/python-semver/python-semver/issues/%s", "#"),
-    "pr": ("https://github.com/python-semver/python-semver/pull/%s", "PR #"),
+    "gh": ("https://github.com/python-semver/python-semver/issues/%s", "#%s"),
+    "pr": ("https://github.com/python-semver/python-semver/pull/%s", "PR #%s"),
 }
 
 # -- Options for HTML output ----------------------------------------------
--- a/docs/migration/migratetosemver3.rst
+++ b/docs/migration/migratetosemver3.rst
@@ -3,8 +3,9 @@
 Migrating from semver2 to semver3
 =================================
 
-This document describes the visible differences for
+This section describes the visible differences for
 users and how your code stays compatible for semver3.
+Some changes are backward incompatible.
 
 Although the development team tries to make the transition
 to semver3 as smooth as possible, at some point change
@@ -34,9 +35,16 @@ Use semver.cli instead of semver
 --------------------------------
 
 All functions related to CLI parsing are moved to :mod:`semver.cli`.
-If you are such functions, like :func:`semver.cmd_bump <semver.cli.cmd_bump>`,
+If you need such functions, like :func:`semver.cmd_bump <semver.cli.cmd_bump>`,
 import it from :mod:`semver.cli` in the future:
 
 .. code-block:: python
 
    from semver.cli import cmd_bump
+
+
+Use semver.Version.is_valid instead of semver.Version.isvalid
+-------------------------------------------------------------
+
+The pull request :pr:`284` introduced the method :meth:`Version.is_compatible <semver.Version.is_compatible>`. To keep consistency, the development team
+decided to rename the :meth:`isvalid <semver.Version.isvalid>` to :meth:`is_valid <semver.Version.is_valid>`.
--- a/docs/migration/replace-deprecated-functions.rst
+++ b/docs/migration/replace-deprecated-functions.rst
@@ -31,6 +31,11 @@ them with code which is compatible for future versions:
 
   Likewise with the other module level functions.
 
+* :func:`semver.Version.isvalid`
+
+   Replace it with :meth:`semver.Version.is_valid`:
+
+
 * :func:`semver.finalize_version`
 
   Replace it with :func:`semver.Version.finalize_version`:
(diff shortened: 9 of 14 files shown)


src/semver/version.py
```diff
@@ -661,6 +661,35 @@ def isvalid(cls, version: str) -> bool:
...
+
+        :param callee: the callee to check for compatibility
+        :return: True, if caller is compatible callee,
+                 otherwise False
+        """
+        if not isinstance(callee, type(self)):
+            raise TypeError(
+                f"Expected a Version type but got {type(callee)}"
+            )
+        if self._major != callee.major:
```
Lexicality, line 685: ✍️ `elif`?

tests/test_semver.py
```diff
@@ -80,3 +80,18 @@ def test_should_versioninfo_isvalid():
 def test_versioninfo_compare_should_raise_when_passed_invalid_value():
     with pytest.raises(TypeError):
         Version(1, 2, 3).compare(4)
+
+
+@pytest.mark.parametrize(
+    "caller, callee",
```
Lexicality, line 86: 💡This is entirely personal preference but I don't like pytest doing string operations 😶
```suggestion
    ("caller", "callee"),
```

tests/test_semver.py
```diff
@@ -80,3 +80,18 @@ def test_should_versioninfo_isvalid():
...
+    [
+        ((1, 2, 3), (1, 2, 3)),
+        ((1, 2, 3), (1, 2, 4)),
+        ((1, 2, 3), (1, 2, 4)),
```
Lexicality, line 90: ❗ you've got the same test here twice, did you mean 
```suggestion
        ((1, 2, 4), (1, 2, 3)),
```

src/semver/version.py
tomschr: I think, in that case it doesn't matter. Python won't continue after a `raise`. :wink:

tests/test_semver.py
tomschr: That would be inconsistent with the other tests. You are right, but I like consistency.

tomschr:
Thanks @Lexicality for your review! Much appreciated! :+1:

Lexicality:
Probably worth mentioning in the docstring that this comparison ignores patch versions and giving a couple of examples?

tomschr:
> Probably worth mentioning in the docstring that this comparison ignores patch versions and giving a couple of examples?

Yes, also something for the documentation. On my todo list. :slightly_smiling_face:

tomschr:
Hi @Lexicality  (and to whom who are interested):
I worked on the issue and updated this PR. Would you mind having a look again? Especially on the test cases and the implementation.

Thank you very much!

rafalkrupinski:
@tomschr great work!

Tests comparing a release with a pre-release and major-0 versions (initial development) are missing;
these should all return false unless versions are equal.

src/semver/version.py
```diff
@@ -661,6 +661,37 @@ def isvalid(cls, version: str) -> bool:
...
+            raise TypeError(
+                f"Expected a Version type but got {type(new)}"
+            )
+        if self._major != new.major:
+            return False
+        if new.minor < self._minor:
+            return False
+        if self._prerelease != new.prerelease:
+            return False
+        return True
```
tlaferriere, line 693:
```suggestion
        return self._major == new.major and new.minor > self._minor and self.prerelease == new.prerelease
```
Here is a suggestion for a slightly cleaner check (IMO). I wrote this in the Github UI so you may want to run black on it to format it properly since it is a rather long statement.

tomschr:
> great work!

Thanks. :blush: 

> Tests comparing a release with a pre-release and major-0 versions (initial development) are missing; these should all return false unless versions are equal.

Ahh yes, right. I'm not completely sure if I understood your sentence. Do you mean something like this:

    >>> ver = Version(0, 0, 0, "rc.1")
    >>> ver.is_compatible(Version(...))  # ???

src/semver/version.py
tomschr: Yes, I was thinking about the same. My approach was to make it obvious first and optimize later. :wink: 

However, when I add your code, I get these errors:

```
$ tox -e checks,py37
[...]
tests/test_semver.py ...............FFFF.......
[...]
______________________________________ [doctest] check-compatible-semver-version.rst _______________________________________
020 
021     # Two different majors:
022     >>> a = Version(1, 1, 1) # This is "A"
023     >>> b = Version(2, 0, 0) # This is "B"
024     >>> a.is_compatible(b)
025     False
026 
027     # The same two majors and minors:
028     >>> b = Version(1, 1, 0) # This is another "B"
029     >>> a.is_compatible(b)
Differences (ndiff with -expected +actual):
    - True
    + False

/home/tom/repos/GH/tomschr/python-semver/docs/usage/check-compatible-semver-version.rst:29: DocTestFailure
================================================= short test summary info ==================================================
FAILED tests/test_semver.py::test_should_succeed_compatible_match[old0-new0] - assert False
FAILED tests/test_semver.py::test_should_succeed_compatible_match[old1-new1] - assert False
FAILED tests/test_semver.py::test_should_succeed_compatible_match[old2-new2] - assert False
FAILED tests/test_semver.py::test_should_succeed_compatible_match[old3-new3] - AssertionError: assert False
FAILED docs/usage/check-compatible-semver-version.rst::check-compatible-semver-version.rst
============================================== 5 failed, 334 passed in 0.55s ===============================================
E
```

tlaferriere: Oops I didn't invert the minor condition properly 😆
```suggestion
        return self._major == new.major and new.minor >= self._minor and self.prerelease == new.prerelease
```

tomschr: Haha, no problem. :slightly_smiling_face: I've corrected it. Could you also answer the question in https://github.com/python-semver/python-semver/pull/368#issuecomment-1320319059 ? Many thanks!

tomschr:
@rafalkrupinski Thanks Raphael for your help. :+1:  I've amended your examples in `test_semver.py`. There is one question left.

If I read item 9 from the Semver spec correct, the other way around, it should be True, right?

```python
Version(1,0,0, 'rc1.').is_compatible(Version(1,0,0))
# returns True
```

Because, `1.0.0-rc1.` < `1.0.0`. Do I miss something?

rafalkrupinski:
> A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version.

That's `False` to me.

tomschr:
> > A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version.
> 
> That's `False` to me.

Ok, that's it! Thanks. I've corrected the tests now. Would you go through it and have a final look?

tomschr:
Oh, one thought that just came into my mind: do we want to make the `is_compatible` visible also the our CLI version?

rafalkrupinski:
@tomschr 
I've edited my previous answer to mention ordering.
Also:
https://github.com/python-semver/python-semver/pull/368/files#diff-d6e8c013b9e7be774edf21bed5411e65f2a2bf6b43110b67e6c4dccdcab23791R683
https://github.com/python-semver/python-semver/pull/368/files#diff-d6e8c013b9e7be774edf21bed5411e65f2a2bf6b43110b67e6c4dccdcab23791R689

tomschr:
> I've edited my previous answer to mention ordering.

Thanks

> And BTW, clearly pre-release is always < release

Yes, that was also my take on this. :smile: Probably I was confused between ordering and compatibility.

You've mentioned these two lines:

* `if not isinstance(new, type(self)):`
   Not sure what's wrong with that. We check first, if we have compatible Version classes. If not, we don't even try to convert them.

* `if (0 == self.major == new.major) and (self[:3] != new[:3]):`
   Also not sure what's wrong either. We need a special if-clause to catch major-0 releases.

Maybe could you elaborate on the two lines, please? Many thanks!

rafalkrupinski:
I don't know how to link comments on code...



> Why checking against type(self)?
> Are you're expecting users to subclass Version? But then user couldn't compare custom Version with this one
> 
> ```python
> class CustomVersion(Version):
>     ...
> 
> CustomVersion(1,0,0).is_compatible(Version(1,0,0)) # raises TypeError
> ```
> 
> Is that the intention?

> Doesn't [:3] make 0.1.0 and 0.1.0-rc1 return True?

tomschr:
> > Why checking against type(self)?
> > Are you're expecting users to subclass Version? But then user couldn't compare custom Version with this one
> > ```python
> > class CustomVersion(Version):
> >     ...
> > 
> > CustomVersion(1,0,0).is_compatible(Version(1,0,0)) # raises TypeError
> > ```
>
> Is that the intention?

Well, not really. My intention was to check against incompatible types, or "anything that is not a Version" is incompatible. Of course, a derived class from Version should be allowed.

Ahh, I think it's this expression:

```
if not isinstance(new, type(self)):
```

it should be:

```
if not isinstance(new, Version):
```

Not sure why I came up with this nonsense. :smile:  I will amend the tests.

 
> > Doesn't [:3] make 0.1.0 and 0.1.0-rc1 return True?

Ahh, right. It gives back major, minor, and patch. Not the pre-release. Should be `self[:4] != new[:4]`.

Both parts are now added (the first part contains also an additional test case).

tomschr:
@Lexicality, @tlaferriere anything that should be changed or improved from your side? Thanks!

rafalkrupinski:
https://github.com/python-semver/python-semver/pull/368/files#diff-20b056eb54729a600ccb91efc5249db4fb816c4d964bdbc7963641694ee0439aR7

Not sure about the wording of this section. 

- All conditions must be met, but bullet 2 is repeated as part of bullet 3.
- bullet 3 - minor is lower _or equal_
- bullet 4  - both are present and equal _or both are absent_.

I'd write:

`a.is_compatible(b) is True` when either is true:
- both versions are identical, or
- both are releases, have equal major and minor versions, or
- both are releases, have equal majors and b's minor version is higher then a's

---
I'm missing instruction on how to use it in real life.

E.g.:
- when package is reading a versioned JSON file, should it check `package_version.is_compatible(file_version)` or the other way around?
- when an executable works with library/API in version X, would it also work with version Y? Should it check X.is_compatible(Y)?

Basically document the caller/callee description that we ditched earlier.
(We want `caller.is_compatible(callee)` and in the examples `file_version.is_compatible(consumer_version)` and yes)

And BTW, I think this best describes the method and it makes its behaviour clear:

>  `a.is_compatible(b)` checks if the _change_ from version "a" to version "b" _is (declared to be) compatible_ according to SemVer rules.

---
Code examples
- I find it harder to read without a=Version... for every example
- it would be nice to include reverse call b.is_compatible(a); extra points if you know why I skipped it from the last example :P
- missing example with different minor, resulting with True
- two examples with pre-release are essentially the same
- is that dot in 'rc1.' on purpose?

```python
    # Two different majors:
    >>> a = Version(1, 1, 1) 
    >>> b = Version(2, 0, 0) 
    >>> a.is_compatible(b)
    False
    >>> b.is_compatible(a)
    False

    # Two different minors:
    >>> a = Version(1, 1, 0) 
    >>> b = Version(1, 0, 0)
    >>> a.is_compatible(b)
    False
    >>> b.is_compatible(a)
    True

    # The same two majors and minors:
    >>> a = Version(1, 1, 1) 
    >>> b = Version(1, 1, 0) 
    >>> a.is_compatible(b)
    True
    >>> b.is_compatible(a)
    True

    # Release and pre-release:
    >>> a = Version(1, 1, 1)
    >>> b = Version(1,0,0,'rc1')
    >>> a.is_compatible(b)
    False
    >>> b.is_compatible(a)
    False

    # Different pre-releases:
    >>> a = Version(1, 0, 0, 'rc1')
    >>> b = Version(1, 0, 0, 'rc2')
    >>> a.is_compatible(b)
    False
    >>> b.is_compatible(a)
    False

    # Identical pre-releases
    >>> a = Version(1, 0, 0,'rc1')
    >>> b = Version(1, 0, 0,'rc1')
    >>> a.is_compatible(b)
    True
```

tomschr:
@rafalkrupinski Thanks a lot Raphael. This is gold! :1st_place_medal: 

You are right with the examples. I used your code and added it there. :+1:  Regarding the real-life examples, I agree. I'm still trying to find good explanation and wording. I will amend it later.

I have some questions/comments regarding your description of the `a.is_compatible(b) is True`:

* Using the expression `a.is_compatible(b) is True` is very useful. :+1:  I like that!
* "both versions are identical".
   You mean both versions are **equal**, right? :wink: Sorry, but I think we should be consistent here as "identical" smells like we are comparing identity with the `is` operator instead of equality with `==`. You probably meant the latter.
* What do you mean by "both are releases"?
   I suppose something like both versions contain prereleases which are equal?
* "their majors are equal and higher than 0 and minor versions"
   That sounds a bit strange to me. If I got it right, you mean that the minors of both versions are also equal and higher than 0. Practically the same conditions than the major's, right?
   I would recommend to dissect it into different main clauses to make it easier to read.
* The order
   In your second and third item, you start with "both are releases". Hmn, that means you're putting the cart before the horse? :wink: 
   Usually we interpret a version from left to right (first major, then minor, then patch, etc.). As such, I would mention the major condition(s) first before explaining the other parts.
  
When I apply these changes from above, this would lead to:

---
The expression ``a.is_compatible(b) is True`` when either is true:

* both versions are equal, or
* both majors are equal and higher than 0. The same applies for both minor parts. Both pre-releases are equal, or
* both majors are equal and higher than 0. The minor of `b`'s minor version is higher then `a`'s. Both pre-releases are equal.
---

I know, it's longer. However, I think, it's clearer.


Many thanks and much appreciated your feedback! :+1:

rafalkrupinski:
1. You should see people righting poems in Perl 4 ;)
2. Of course, equal. I used `identical` because definitions of equality can vary.
3. By release I mean not a pre-release. Pre-releases need to be equal to be compatible so they fall in the rule 1. The two other rules apply only to release versions. Same with major-zero
4.

> Practically the same conditions than the major's, right?

Minor parts don't have to be higher than zero to be compatible.

5. OK, so maybe it's clearer when it's less condensed. SemVer explains it in 9 points :)
(Although, they also define some ordering rules)

> - they are equal, or
> - all of:
>     - their major parts are higher than 0 and equal, and
>     - neither of them is a pre-release, and
>     - b's minor is greater than or equal a's

tomschr:
> 1. You should see people writing poems in Perl 4 ;)

:laughing: Luckily, we don't have to follow Perl. :wink: 


>     3. By release I mean not a pre-release. Pre-releases need to be equal to be compatible so they fall in the rule 1. The two other rules apply only to release versions. Same with major-zero

Ah, I see. Hmn. I think, the term "release" could be mistaken by "pre-release" (as I did). I would like to avoid this. That's why I wrote in all of the documentation about "versions", not "releases".


> > Practically the same conditions than the major's, right?
> 
> Minor parts don't have to be higher than zero to be compatible.

I know, but for your item ""their majors are equal and higher than 0 and minor versions"" I interpret it as such. :slightly_smiling_face:

rafalkrupinski:
> > 3. By release I mean not a pre-release. Pre-releases need to be equal to be compatible so they fall in the rule 1. The two other rules apply only to release versions. Same with major-zero
> 
> Ah, I see. Hmn. I think, the term "release" could be mistaken by "pre-release" (as I did). I would like to avoid this. That's why I wrote in all of the documentation about "versions", not "releases".

Perhaps you're just so immersed in this project and the lingo, that you've just read what you expected to read ;)

Releases and pre-releases have different compatibility rules, so you can't just replace 'release' with 'version', unless you write 'version other than pre-release', but to me it's just indirect way of writing 'release' 🤷🏻. Back to square one.

> > > Practically the same conditions than the major's, right?
> > Minor parts don't have to be higher than zero to be compatible.
> I know, but for your item ""their majors are equal and higher than 0 and minor versions"" I interpret it as such. 🙂

Missing coma perhaps? Don't let my bad grammar put you off track ;)

Hopefully this approach is clearer: https://github.com/python-semver/python-semver/pull/368#issuecomment-1362674770

tomschr:
Sorry, it took me some time to revisit this. I'd like to merge it to create the final 3 release if nobody objects.

Thanks for all your help!

tomschr:
As nobody seemed to object, I'm merging it. Thank you!

Issue #374, opened 2022-11-26
Update Towncrier config in pyproject.toml


Nagidal:
# Situation
Current towncrier configuration uses a deprecated format (see Towncrier's [configuration reference](https://towncrier.readthedocs.io/en/stable/configuration.html#custom-fragment-types))

```toml
  [[tool.towncrier.type]]
  directory = "trivial"
  name = "Trivial/Internal Changes"
  showcontent = true
```

# Expected Behavior
Use the new format:
```toml
  [tool.towncrier.fragment.trivial]
  name = "Trivial/Internal Changes"
  showcontent = true
```
(array of toml tables is [deprecated](https://towncrier.readthedocs.io/en/stable/configuration.html#deprecated-defining-custom-fragment-types-with-an-array-of-toml-tables))

tomschr:
Thank you very much, @Nagidal! I wasn't aware of that. I take care of updating the configuration for Towncrier.

Issue #372, opened 2022-11-17
Remove support for Python 3.6?


tomschr:
# Situation
Python 3.6 has reached its end of life and isn't available for download in https://www.python.org/downloads/ anymore. At the time of writing (Nov 2022), the lowest version is 3.7.

# Possible Solution/Idea
Remove support for Python 3.6 for semver3. Alternative would be to remove it in semver3.1.
After we get a clearer picture from the poll below we will act accordingly.

To remove support for Python 3.6, we need to adapt the following files:

* [x]  `setup.cfg` in `classifiers` and  `python_requires`
* [x] `.github/workflows/python-testing.yml`
* [x] `pyproject.toml`
* [x] `tox.ini`
* [x] `README.rst`
* [x] `CONTRIBUTING.rst`
* [x] Review documentation (examples that point to Python 3.6 needs an update)
* [x] Is `setup.py` still needed? Can it be removed? Yes, was only needed for compatibility reasons with Python 3.6.

# Additional context
* See poll in https://github.com/python-semver/python-semver/discussions/371#discussion-4575292
* [PEP 0494](https://peps.python.org/pep-0494/)

---

Hi @tlaferriere, any opinion? :slightly_smiling_face:

tlaferriere:
I think we should drop support for python 3.6 in the next release.
Dropping support for python 3.6 does not mean all the previous versions published on PyPI will be removed and unavailable, so the impact for users on python 3.6 will not be felt if they do not upgrade the package indiscriminately. Also to prevent accidental upgrades, I think we can set the minimum python version in the `setup.(py|cfg)`and that should prevent python 3.6 users from installing it. Usually, users of eol python versions generally do not actively develop their applications, therefore would not want to use new features just for the sake of it.

Otherwise, I think dropping support for an eol python version on its own would not warrant a major version release for two reasons. First, it is not a public API change, which is what semver compatibility is all about. This would make it a patch release, since it is an implementation change (that may break dependencies of they depended on the"buggy" implementation). Second, there are other mechanisms to preserve compatibility on the python interpreter level in this context.

For these reasons, limited impact on the community and the fact that it is not a public API change, I think we should not think twice about dropping support for eol python versions, but make sure that users of eol python versions can't upgrade by using the appropriate configuration for our PyPI deployment.

tomschr:
Thanks Thomas for your opinions. :+1: I had similar thoughts, but you summed it up perfectly.

[...]

PR #376, opened 2022-11-26
Hatchbuild

### LINKED ISSUE #373: Use hatchling build backend
# Situation
python-semver currently uses the setuptools build backend

# Possible Solution/Idea
hatchling seems to be the recommended build backend now. (Tutorial [uses it](https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml) by default.)

# Additional context
One of the benefits of switching to [hatch](https://hatch.pypa.io) (which uses hatchling), besides a nice venv management, would be the possibility of using [hatch-semver](https://github.com/Nagidal/hatch-semver) plugin to semantically version this project.

Nagidal:
Proposal to fix #373 (Use hatchling build backend) and #374 (Update Towncrier config in pyproject.toml)

I have moved everything except for flake8 settings and pycodestyle settings from `setup.cfg` to `pyproject.toml`

I have prepared these scripts:

- `style:fmt` runs Black and isort (all files now have imports sorted by isort)
- `style:lint` runs flake8 and pycodestyle
- `docs:build` runs pdoc3 to build API documentation in the `api-docs` directory (added to `.gitignore`
- `docs:serve` runs pdoc3 local server on `localhost:8080` to observe API docs updated live as you update docstrings. However, this somehow does not work in this project and I don't know exactly how to solve the error. 
- `test:cov` runs pytest-cov
- `test:no-cov` runs pytest with `--no-cov`
- `test:tox_test` runs tox

Scripts can be run with `hatch run`, e.g. `hatch.run style:fmt`. All the test scripts fail because of `python-semver\tests\coerce.py` not being a proper Python file (invalid syntax). I don't know what purpose of this file is, so I did not touch it. It makes tests fail even if you run then natively with `tox`, so repairing this is beyond the scope of this PR, methinks.

I changed the classifier `Environment :: Web Environment` to `Environment :: Console` as I think it's more fitting.

I also removed Python 3.6 from black's matxix in anticipation of #372 (Remove support of Python 3.6)


Nagidal:
Ok, I see I should have not mixed in the issues #374 and #372 in here. I'll prepare another PR

PR #396, opened 2023-02-27
Fix return type of classmethod parse

def-:
So that calling parse on a derived class will show correct type of derived class


tomschr:
Thank you very much @def-! :+1: 
I'm not so familiar with type hints, so this is very helpful.

The only minor change that I did was to move the line with `T = TypeVar("T", bound="Version")` up so all definitions are together.

Many thanks!

PR #405, opened 2023-04-03
Remove incorrect dependencies from build-system section of pyproject.toml

mgorny:
Remove `wheel` and `build` dependencies from `pyproject.toml`. The former is added automatically by setuptools and is incorrect for sdist, the latter is frontend rather than backend.

(described in detail in commit messages)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -99,10 +99,11 @@ commands = make -C docs man
 description = Prepare for TestPyPI
 basepython = python3
 deps =
-    wheel
     twine
-    # PEP 517 build frontend
-    build
+allowlist_externals =
+    # PEP 517 build frontend, automatically installed by setuptools
+    pyproject-build
+    twine
 commands =
     # Same as python3 -m build
     pyproject-build
--- a/tox.ini
+++ b/tox.ini
@@ -100,10 +100,7 @@ description = Prepare for TestPyPI
 basepython = python3
 deps =
     twine
-allowlist_externals =
-    # PEP 517 build frontend, automatically installed by setuptools
-    pyproject-build
-    twine
+    build
 commands =
     # Same as python3 -m build
     pyproject-build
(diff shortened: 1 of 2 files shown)


tomschr:
@mgorny Thanks for your contribution! Much appreciated! :+1:

This was a remnant from Python 3.6. I've also updated `tox.ini` which contained these dependencies.

If you are okay with it, I can merge it.

tox.ini
```diff
@@ -99,10 +99,11 @@ commands = make -C docs man
...
 basepython = python3
 deps =
-    wheel
     twine
-    # PEP 517 build frontend
-    build
+allowlist_externals =
+    # PEP 517 build frontend, automatically installed by setuptools
+    pyproject-build
+    twine
```
mgorny, line 106:
```suggestion
    build
```

No, that's a bit incorrect. `wheel` is installed automatically inside the build environment, so that one is unnecessary.

`build` is necessary if you're calling it explicitly. You don't list it in `pyproject.toml` because `pyproject.toml` decides what's used by `build`. However, you need to list it in `tox.ini` if you intend to call `pyproject-build` from `tox`.

So basically, tox installs `build` and calls `pyproject-build`. Then `build` installs `setuptools` based on `pyproject.toml`, and calls its backend, and setuptools' backend installs `wheel` automatically.

mgorny: Hmm, I'm not sure if GitHub got it right — I was trying to readd `build` into `deps`, not `allowlist_externals`. The latter is for system tools that you're calling; however, since `build` is a pure Python package, it's better to pull it via `deps` than require the system to have one.

tomschr: Oh, I see. I agree, `build` should be included in `deps`.

I've added another commit. Interestingly, if you add `build` into `deps`, you don't need the `allowlist_externals` anymore.

I've tried it and it works for me. Can you confirm?

mgorny: Yes, this version looks correct. Thanks!

tomschr: Perfect! I'll squash merge it to master.

Many thanks for your contributions! :+1:

PR #408, opened 2023-04-22
Improve GitHub Action

tomschr:
* Set timeout-minutes time to 15min
* Add on.push.path on.pull_request.path
* Raise Python 3.7 -> 3.8
* Use cache for actions/setup-python@v3

Ideas taken from https://ashishb.net/tech/common-pitfalls-of-github-actions/


tomschr:
Let's try it.

PR #418, opened 2023-07-06
Update doc string for python 3.12

dschwoerer:
Resolves test failure with python3.12:

Failed [build](https://copr.fedorainfracloud.org/coprs/davidsch/testing/build/6147136/) and [log](https://download.copr.fedorainfracloud.org/results/davidsch/testing/fedora-rawhide-x86_64/06147136-python-semver/builder-live.log.gz)

Fixed [build](https://copr.fedorainfracloud.org/coprs/build/6147216) and [log](https://download.copr.fedorainfracloud.org/results/davidsch/testing/fedora-rawhide-x86_64/06147216-python-semver/builder-live.log.gz)

### DIFF
--- a/docs/usage/convert-version-into-different-types.rst
+++ b/docs/usage/convert-version-into-different-types.rst
@@ -17,7 +17,7 @@ It is possible to convert a :class:`~semver.version.Version` instance:
 
     >>> v = Version(major=3, minor=4, patch=5)
     >>> v.to_dict()
-    OrderedDict([('major', 3), ('minor', 4), ('patch', 5), ('prerelease', None), ('build', None)])
+    OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None})
 
 * Into a tuple with :meth:`~semver.version.Version.to_tuple`::
 
--- a/docs/usage/create-a-version.rst
+++ b/docs/usage/create-a-version.rst
@@ -90,7 +90,7 @@ Depending on your use case, the following methods are available:
   To access individual parts, you can use the function :func:`semver.parse`::
 
     >>> semver.parse("3.4.5-pre.2+build.4")
-    OrderedDict([('major', 3), ('minor', 4), ('patch', 5), ('prerelease', 'pre.2'), ('build', 'build.4')])
+    OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': 'pre.2', 'build': 'build.4'})
 
   If you pass an invalid version string you will get a :py:exc:`ValueError`::
 
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -228,8 +228,7 @@ class Version:
           ``patch``, ``prerelease``, and ``build``.
 
         >>> semver.Version(3, 2, 1).to_dict()
-        OrderedDict([('major', 3), ('minor', 2), ('patch', 1), \
-('prerelease', None), ('build', None)])
+        OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None})
         """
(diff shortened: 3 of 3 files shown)


tomschr:
Hi @dschwoerer, thanks for your report.

This is a bit confusing. I've tried a Python shell session:

```python
$ python3.12 --version
Python 3.12.0b2 (main, Jul  7 2023, 07:49:42) [GCC 7.5.0] on linux
$ python3.12
>>> from collections import OrderedDict
>>> OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None})
OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None})

>>> OrderedDict([('major', 3), ('minor', 4), ('patch', 5), ('prerelease', None), ('build', None)])
OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None})
```

Both syntax variants are possible with `OrderedDict`.  I'm just not sure why it doesn't work in the doctest. :thinking: 

Any idea?

tomschr:
After I've fixed the long line issue, it seems the GitHub Actions are not so happy with your change. :wink: 

Your "fix" may work for Python 3.12, but breaks the other versions. I'm sorry, but I can't accept this fix as it is.

We need to investigate further.

tomschr:
Ok, I digged deeper and there is a change between 3.12 and the other Python versions. It's in the `__repr__` method.

In Python 3.12 you have this code (see [source](https://github.com/python/cpython/blob/3.12/Lib/collections/__init__.py#L271-L275)):

```python
@_recursive_repr()
def __repr__(self):
  'od.__repr__() <==> repr(od)'
  if not self:
      return '%s()' % (self.__class__.__name__,)
  return '%s(%r)' % (self.__class__.__name__, dict(self.items()))
```

whereas in Python 3.11 you have this (see [source](https://github.com/python/cpython/blob/3.11/Lib/collections/__init__.py#L266-L270)):

```python
@_recursive_repr()
def __repr__(self):
  'od.__repr__() <==> repr(od)'
  if not self:
      return '%s()' % (self.__class__.__name__,)
  return '%s(%r)' % (self.__class__.__name__, list(self.items()))
```

The last line is the culprit.

Seems we need to change the doctest somehow which does not relate to `OrderedDict`. I have something like this in mind:

```
>>> list(OrderedDict({'major': 3, 'minor': 4, 'patch': 5, 'prerelease': None, 'build': None}).items())
[('major', 3), ('minor', 4), ('patch', 5), ('prerelease', None), ('build', None)]
```

dschwoerer:
I do not know doc tests, so I only managed to get the fix for 3.12

As you claim to support 3.12 on pypi, I thought you may be interested :-)

It may be best to always return a dict?
They are sorted since python3.6. I don't think you need OrderedDicts, the benefit OrderedDicts provide are not likely useful.

Do you want me to change it to return dict? That way the doc test would not need to be conditionalised ...

Of course, using to list or something could also work ...

tomschr:
Thanks for your answer! I think I found a solution to this problem which I committed to 686ce67.

> It may be best to always return a dict?

That could indeed an option. I haven't thought about that. :smile:  The OrderedDict was introduced due to Python 3.6 as the dict wasn't ordered. As we don't support this version anymore, we might use a simple dict.

> Do you want me to change it to return dict? That way the doc test would not need to be conditionalised ...

Yes, please! That would be great! :+1: You need to adapt these files:

* `docs/usage/convert-version-into-different-types.rst`
* `docs/usage/create-a-version.rst`
* `src/semver/version.py`

Look for my last commit to see which lines.

tomschr:
This PR can be closed as #419 is merged now. Thanks!

PR #419, opened 2023-07-12
Return dict instead of OrderedDict

dschwoerer:
dict is guaranteed to preserver order in all supported versions

Replaces #418


tomschr:
@dschwoerer Thanks David for your contribution! :+1:

PR #424, opened 2023-07-27
Only run GHA if mentioned files are modified

tomschr:


tomschr:
Close it for the time being

PR #427, opened 2023-09-28
Fix #426: call subclass when deriving from Version

### LINKED ISSUE #426: Replace method not compatible with SubClasses
# Situation

Using the demonstration subclass from https://python-semver.readthedocs.io/en/latest/advanced/create-subclasses-from-version.html#creating-subclasses-from-version, the bump methods will create a new instance of the same SubClass.

When using the `replace` method, an instance of the `Version` class is created and not the subclass.

# To Reproduce

1. Create a subclass as shown in the document above.
2. `version = SemVerWithVPrefix.parse("v1.1.0")`
3. `dev_version = version.replace(prerelease="dev.0")`
4. Print `str(dev_version)`
5. This results in `1.1.0-dev.0` - dev_version is of type `Version` not `SemVerWithVPrefix`.

# Expected Behavior

I expect `v1.1.0-dev.0` and `dev_version` to be an instance of SemVerWithVPrefix (or my class).

# Environment
- OS: [e.g. Linux, MacOS, Windows, ...] - Python 3.11 container - will be Debian.
- Python version [e.g. 3.6, 3.7, ...] - Python 3.11.5
- Version of semver library [e.g. 3.0.0] - semver==3.0.1

# Additional context

I suggest changing this the same way that the bump methods work:
Add this near the top: `cls = type(self)`
And near the bottom, replace `return Version(**version)` with `return cls(**version)`.

tomschr:
This fixes PR #426 

When using the replace method, an instance of the Version class is created and not the subclass. This commit fixes this behavior.

@dannystaple: Could you have a look, please? :slightly_smiling_face: 

(The CI has currently problems. I'm currently investigating this.)

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -60,7 +60,7 @@ jobs:
           cache: 'pip'
       - name: Install dependencies
         run: |
-          python3 -m pip install --upgrade pip setuptools setuptools-scm
+          python3 -m pip install --upgrade pip setuptools>60 setuptools-scm>=60
           pip install tox tox-gh-actions
       - name: Check
         run: |
--- a/changelog.d/426.bugfix.rst
+++ b/changelog.d/426.bugfix.rst
@@ -1,0 +1,2 @@
+Fix :meth:`~semver.version.Version.replace` method to use the derived class
+of an instance instead of :class:`~semver.version.Version` class.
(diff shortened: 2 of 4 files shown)


dannystaple (approved):

dannystaple:
This looks good to me. I totally get investigating CI.

tomschr:
@dannystaple I could fix the CI problem. Thank you very much for your contribution. :+1:

PR #453, opened 2024-12-13
Fix comparison with subclasses

viccie30:
The check in `_comparator` does not match the check in `Version.compare`, breaking comparison with subclasses.


tomschr:
Thanks a lot @viccie30, good catch! :+1:

PR #472, opened 2026-07-29
fix(next_version): reset prerelease when token changes (#339)

### LINKED ISSUE #339: increasing next_version 'prerelease' with different prerelease_token
# Situation
I am not shure if this should be seen as a bug or an enhancement. I have observed this behaviour in version 2.13 of semver.

If I have a version 1.2.3 

`ver = semver.VersionInfo.parse('1.2.3')`

and would like to get the next dev Version, 
I will get the version 1.2.4-dev.1 by calling

`ver_dev1 = ver.next_version('prerelease', prerelease_token='dev')`

If I would like to get the next release candidate, I would call

`ver_rc1 = ver_dev1.next_version('prerelease', prerelease_token='rc')`

which should be 1.2.4-rc.1

but the result is 1.2.4-dev.2

# To Reproduce
```
import semver
ver = semver.VersionInfo.parse('1.2.3')
ver_dev1 = ver.next_version('prerelease', prerelease_token='dev')
ver_rc1 = ver_dev1.next_version('prerelease', prerelease_token='rc')
print(ver_rc1)
```
The result is

`1.2.4-dev.2`

# Expected Behavior

I would assume, that according to semantic versioning the result should be

`1.2.4-rc.1`

# Environment
- OS: Linux
- Python version 3.6.13
- Version of semver library 2.13

Mukller:
## Problem

Fixes #339.

`next_version('prerelease', prerelease_token='rc')` ignores the token when the version already has a prerelease with a **different** token:

```python
import semver
ver  = semver.Version.parse('1.2.3')
dev1 = ver.next_version('prerelease', prerelease_token='dev')  # 1.2.4-dev.1
rc1  = dev1.next_version('prerelease', prerelease_token='rc')  # 1.2.4-dev.2 (bug!)
# expected: 1.2.4-rc.1
```

## Root Cause

`next_version` delegates to `bump_prerelease(token, bump_when_empty=True)`. Inside `bump_prerelease`, when a prerelease already exists the `token` argument is silently ignored:

```python
if self._prerelease is not None:
    # token is completely ignored here:
    prerelease = cls._increment_prerelease(self._prerelease)  # dev.1 -> dev.2
else:
    prerelease = str(token) + '.1'  # token used only for brand-new prerelease
```

## Fix

Before delegating to `bump_prerelease`, `next_version` now checks whether the existing prerelease token matches the requested one. When they differ, the prerelease is reset to `{prerelease_token}.1`:

```diff
 else:
+    if version.prerelease is not None and prerelease_token:
+        current_token = version.prerelease.split('.')[0]
+        if current_token != prerelease_token:
+            return version.replace(prerelease=f'{prerelease_token}.1', build=None)
     return version.bump_prerelease(prerelease_token, bump_when_empty=True)
```

## Behaviour Table After Fix

| Input | `prerelease_token` | Result |
|---|---|---|
| `1.2.3` | `'rc'` | `1.2.4-rc.1` (unchanged) |
| `1.2.4-rc.1` | `'rc'` | `1.2.4-rc.2` (unchanged) |
| `1.2.4-dev.1` | `'rc'` | `1.2.4-rc.1` (**fixed**) |
| `1.2.4-dev.3` | `'dev'` | `1.2.4-dev.4` (unchanged) |

The patch version is preserved when changing from one prerelease token to another (the version is already in its bump-patch prerelease cycle).

### DIFF
--- a/docs/usage/increase-parts-of-a-version_prereleases.rst
+++ b/docs/usage/increase-parts-of-a-version_prereleases.rst
@@ -15,7 +15,7 @@ would perhaps a better fit.
 
     >>> v = Version.parse("3.4.5-pre.2+build.4")
     >>> str(v.next_version(part="prerelease"))
-    '3.4.5-pre.3'
+    '3.4.5-rc.1'
     >>> str(Version.parse("3.4.5-pre.2+build.4").next_version(part="patch"))
     '3.4.5'
     >>> str(Version.parse("3.4.5+build.4").next_version(part="patch"))
--- a/src/semver/version.py
+++ b/src/semver/version.py
@@ -482,6 +482,10 @@ build='build.10')
         if part in cls.NAMES[:3]:
             return getattr(version, "bump_" + part)()
         else:
+            if version.prerelease is not None and prerelease_token:
+                current_token = version.prerelease.split(".")[0]
+                if current_token != prerelease_token:
+                    return version.replace(prerelease=f"{prerelease_token}.1", build=None)
             return version.bump_prerelease(prerelease_token, bump_when_empty=True)
 
     @_comparator


Mukller (commented):
## Code Review

### Bug Verification

Reproducible on current master:

```python
import semver
v = semver.Version.parse('1.2.4-dev.1')
print(v.next_version('prerelease', prerelease_token='rc'))
# Current master: 1.2.4-dev.2  (bug)
# After patch:    1.2.4-rc.1   (correct)
```

Root cause: `bump_prerelease` ignores its `token` parameter when `self._prerelease is not None`, passing only the existing string to `_increment_prerelease`. The token is only used in the `else` branch (no existing prerelease).

### Fix Correctness

The 4-line guard in `next_version` handles the token-change case entirely in one place, without touching `bump_prerelease`. This is correct because:

- Only `next_version` has the concept of a `prerelease_token` in its API —   `bump_prerelease` documents that it ignores the token for an existing prerelease   (it just bumps the counter).
- `version.replace(prerelease=f'{prerelease_token}.1', build=None)` keeps   the major/minor/patch unchanged — correct, since we are staying in the   same patch-bump cycle.

### Edge Cases

| Scenario | Before | After |
|---|---|---|
| No prerelease | `1.2.4-rc.1` ✓ | `1.2.4-rc.1` ✓ (unchanged) |
| Same token | `1.2.4-rc.2` ✓ | `1.2.4-rc.2` ✓ (unchanged) |
| Different token | `1.2.4-dev.2` ✗ | `1.2.4-rc.1` ✓ (fixed) |
| Empty token `''` | falls to `bump_prerelease` | falls to `bump_prerelease` (guard skipped by `if prerelease_token:`) |
| `None` token | falls to `bump_prerelease` | falls to `bump_prerelease` (guard skipped by `if prerelease_token:`) |

### Change Size

4 lines added to one method. No changes to `bump_prerelease` or `_increment_prerelease`.

tomschr:
Thank you @Mukller for your contribution! Much appreciated! :+1:

I fixed a small doctest. I'm going to merge it.

kaeluka:
Hi, @tomschr and @Mukller - i randomly came across this pr after it had been merged.

I noticed that one thing this changes is that a next version can now be considered _before_ the version it is supposed to succeed.

Per SemVer §11 (and _nat_cmp), alphanumeric identifiers compare by ASCII, so this is only "next" when the requested token happens to sort above the existing one:

```python
from semver import Version

# 'c' < 'e'  (ASCII identifier compare)
v = Version.parse("1.2.3-release.1+b.1")
n = v.next_version("prerelease", prerelease_token="rc")
print(v, "--", "rc", "-->", n, "| next > current:", n > v)

# 'd' < 'r'  (ASCII identifier compare)
v = Version.parse("1.2.3-rc+b.1")
n = v.next_version("prerelease", prerelease_token="dev")
print(v, "--", "dev", "-->", n, "| next > current:", n > v)

# numeric id < alphanumeric id (SemVer section 11)
v = Version.parse("1.2.3-2dev.1+b.1")
n = v.next_version("prerelease", prerelease_token="2")
print(v, "--", "2", "-->", n, "| next > current:", n > v)
 ```

I'm not sure you'd consider this a bug or not, it at least seems quite surprising behaviour, especially in the light of:

> Determines next version, preserving natural order.

https://github.com/python-semver/python-semver/blob/0f18aa0b1e0cabba65ee1b506e637faabcdf5d05/src/semver/version.py#L454

I'd love to hear your response to this - sorry if it's a false alarm.

PR #475, opened 2026-08-31
Declare support for Python 3.15

edgarrmondragon:
Final Python 3.15.0 will be released in about one month.


tomschr:
Many thanks, @edgarrmondragon! This is a good addition. :+1:

PR #477, opened 2026-09-07
Ensure equal versions with a numeric prerelease have equal hashes

ckarnell:
### Equal versions can hash differently when a prerelease identifier is numeric

#283 established that equal versions must have equal hashes and fixed it for build metadata. This is the same invariant for the numeric-prerelease case it did not cover.

`__eq__` compares prereleases through `_nat_cmp`, which reads all-digit identifiers as integers, so these two versions are equal:

```python
from semver import Version
a = Version(1, 2, 3, prerelease="1")
b = Version(1, 2, 3, prerelease="01")
a == b              # True
hash(a) == hash(b)  # False
{a, b}              # two elements, not one
{a: "x"}[b]         # KeyError, on a key that compares equal to one present
```

`__hash__` hashed the raw prerelease string, so the two equal versions land on different hashes and break set and dict membership.

### Scope, stated up front

This is not reachable through `Version.parse`: the parser's regex rejects a leading-zero numeric identifier like `01`. It is reachable through the public `Version(...)` constructor and `.replace()`, which store the prerelease without validating it. So the impact is limited to versions built that way, and I did not want to overstate it.

### Fix

Normalize numeric prerelease identifiers in `__hash__` exactly as `_nat_cmp` (and therefore `__eq__`) already do, so equal versions hash equally. Build metadata stays excluded, matching #283.

Added a regression test next to `test_equal_versions_have_equal_hashes`; it fails without the change. The full suite passes.


tomschr:
Thank you for all your efforts!

I was unsure and had to look into the spec. [Section 9](https://semver.org/#spec-item-9) of the Semantic Versioning 2.0.0 specification explicitly prohibits leading zeroes in numeric pre-release identifiers:

> A pre-release version [...] **Numeric identifiers MUST NOT include leading zeroes**

In other words, a Version `1.2.3-01` would violate the spec and must be rejected. If the semver library accepts that (and it does), then that's the real bug, not the hash.

If I'm not mistaken, NPM's semver has the same approach. Comparing `1.2.3-01` and `1.2.3-1` in strict mode fails too because the former cannot be parsed.

As I strictly follow the semver spec, I'm sorry I will have to reject this PR.

However, we can discuss if this library should allow a strict and loose mode similar to NPM. This isn't implemented yet, but would require some deeper refactoring of the code.

ckarnell:
You're right, and thanks for digging into the spec. Validate-at-construction is the correct fix location: Section 9 makes `1.2.3-01` invalid input, so rejecting it at parse and construction is the real fix, and normalizing it in `__hash__` only papers over an object that shouldn't exist in the first place. The inconsistency did surface from a `==`/`hash` contract check (equal objects must hash equally), which should stay a useful regression test whenever you add the identifier validation. Closing this in favor of that approach.

PR #478, opened 2026-09-12
Reject non-ASCII digits and trailing newlines in version parsing

Str0k:
`Version.parse()` currently accepts characters outside the SemVer grammar. For example, `Version.is_valid('1.2.3\n')` is true, and `Version.parse('1\u0662.2.3')` silently converts a mixed ASCII/Arabic-Indic major component to `12.2.3`. Non-ASCII digits can also survive in prerelease identifiers.

The shared regex uses Python's Unicode-aware `\d` and `$`, which may match before a final newline. Compile both strict and optional-component patterns with `re.ASCII` and use `\Z` to require the actual end of the input. These constraints follow the ASCII digit grammar and prerelease requirements in https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions.

This intentionally rejects previously accepted malformed input; callers wanting whitespace tolerance can strip it before parsing. No exploit or CVE is claimed.

Validation, Windows / Python 3.12:

- All 23 new regression cases fail on the original code.
- Without the optional native backend: 385 passed, 1 skipped; 99.1% coverage.
- With fast-semver-rs-backend 0.1.2: 386 passed; 99.7% coverage, including the fallback after native rejection.
- Ruff passes on the new test file. The changed implementation file has the same 43 existing Ruff diagnostics as base `0f18aa0`; mypy has the same existing missing-backend-stub diagnostic on both revisions.
- The two documentation symlinks under `tests/` were temporarily materialized as copies for Windows testing, then restored. They are not part of the diff.

AI assistance: Codex helped implement and test this change. The reproduction and verification results are supplied for maintainer review.


tomschr:
Thank you very much @Str0k for your contribution! This is awesome! :tada: :heart_eyes: 

Many thanks!

PR #479, opened 2026-09-12
Handle malformed collection operands in Version comparisons

Str0k:
Comparing a `Version` with a malformed list, tuple, or dictionary leaks errors from the internal constructor. For example, `Version(1, 0, 0) == []` raises `TypeError` even though equality with unsupported operands should evaluate to `False`.

Convert collection operands in the comparison decorator and return `NotImplemented` when construction raises `TypeError` or `ValueError`. Equality then returns `False`, and ordering raises Python's standard unsupported-comparison `TypeError`. Valid collections and subclass behavior are preserved; `Version.compare()` and string parsing are unchanged.

The patch adds 19 regression cases, updates the documentation's comparison example, and includes `changelog.d/pr479.bugfix.rst` using the actual PR number as documented by `changelog.d/README.rst`.

Validation against base `1e8757912f633684e1a65d9d7c7e2270c3c79928`, reproduced independently in clean checkouts:

| Check | Base with new regressions | Patch |
| --- | --- | --- |
| New regression module | 18 failed, 1 passed | 19 passed |
| Canonical `python -m pytest -q`, including documentation | 18 failed, 387 passed | 405 passed |
| Mypy 2.1.0 | 1 existing diagnostic | Same diagnostic |
| Ruff 0.16.0, complete repository | 159 existing diagnostics | Same set; new test file clean |
| docformatter | Existing formatting differences | Identical differences |

The changelog was added in a separate commit after draft creation. Towncrier validation passed; that commit changes only the reviewed newsfragment, leaving the tested code and documentation unchanged.

AI assistance: implementation and independent code review used Hermes with self-hosted GLM 5.3. Automated validation was reproduced separately. This does not represent a human review.


tomschr:
Thank you for the contribution! :+1:

PR #483, opened 2026-09-13
Reject invalid token arguments in Version.bump_prerelease and Version.bump_build

Str0k:
Calling Version.parse("1.0.0").bump_build("build+meta") or bump_prerelease("rc.") previously returned a Version object whose string representation ("1.0.0+build+meta.1", "1.0.0-rc..1") is not valid SemVer and cannot be round-tripped through Version.parse(). After this fix, both methods validate the dot-separated token against the SemVer identifier grammar used by the parser and raise ValueError for malformed tokens such as "bu/ild", "build+meta", "build..2", "rc." or a numeric prerelease segment with leading zeros ("rc.05"). Valid tokens ("build", "rc.1", "dev-1", "alpha.0") and the documented None/empty-string defaults are unchanged.

The parser (Version.parse/is_valid) enforces the SemVer 2.0.0 identifier grammar, but the bump methods concatenated the raw token string without any check, so the library could produce Version objects it would itself reject on the next parse. The fix adds a small _validate_identifiers classmethod that splits the token on '.' and fullmatches each part against the same identifier regexes the parser template uses for prerelease (numeric identifiers without leading zeros, or alphanumeric identifiers) and build ([0-9A-Za-z-]+) parts, and calls it from the explicit-token branch of both bump methods. This keeps the validation rules in one place and leaves the string parser and all other behavior untouched.

Validation:
- Regression on unchanged base: 12 assertion failure(s); with patch: 22 tests, exit 0.
- Full suite: base 439 tests (exit 1); patch 439 tests (exit 0).
- Repository checks: docformatter: exit 1, documented_bump_contract: exit 0, mypy: exit 1, ruff: exit 1.

AI assistance: implementation and independent review used Hermes with GLM 5.3. Test evidence was reproduced in clean checkouts. This does not represent a human review.
An additional model reviewed the supplied patch and evidence without executing tools.

Changelog: the reviewed bugfix entry is attached using this PR number after draft creation, following changelog.d/README.rst. The PR is marked ready only after the fragment passes Towncrier.


src/semver/version.py
```diff
@@ -129,6 +129,12 @@ class Version:
         _REGEX_TEMPLATE.format(opt_patch="?", opt_minor="?"),
         re.VERBOSE | re.ASCII,
     )
+    #: Regex for a single valid build identifier
+    _BUILD_IDENTIFIER: ClassVar[Pattern[str]] = re.compile(r"[0-9A-Za-z-]+", re.ASCII)
+    #: Regex for a single valid alphanumeric prerelease identifier
+    _PRERELEASE_IDENTIFIER: ClassVar[Pattern[str]] = re.compile(
+        r"0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*", re.ASCII
```
tomschr: These regexes are slightly different from the `_REGEX_TEMPLATE`. That template contains for the `prerelease` part this regex:

```python
""" # 
(?:-(?P<prerelease>
                (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
                (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
))?
```

and for the `build` part this:

```python
(?:\+(?P<build>
                [0-9a-zA-Z-]+
                (?:\.[0-9a-zA-Z-]+)*
            ))?
```

I think that kind of violates the DRY principle. If feasible, I would prefer to have _one_ regex source which is then used int he `_REGEX_TEMPLATE`.

Maybe you can use them and refer to it? As a rough sketch, something like this.

```python
# ....
_BUILD_IDENTIFIER: ClassVar[str] = r"..."

# ...
_PRERELEASE_IDENTIFIER: ClassVar[str] = r"..."

# ...
_REGEX_TEMPLATE: ClassVar[Pattern[str]] = r""  # <= Inside, add {_BUILD_IDENTIFIER} and {_PRERELEASE_IDENTIFIER}
```


Perhaps you need to adjust the regex a bit. But that would help to avoid repeating the regexes.

tomschr (changes requested):
Thank you @Str0k! I have a little wish. What do you think?

tests/test_bump_token_validation.py
```diff
@@ -0,0 +1,95 @@
...
+def test_bump_methods_still_accept_empty_and_none(empty):
+    """Empty string and None are documented and must keep their meaning."""
+    assert str(Version.parse("1.0.0").bump_prerelease(empty)) in (
+        "1.0.0-1",
+        "1.0.0-rc.1",
+    )
+    assert str(Version.parse("1.0.0").bump_build(empty)) in (
+        "1.0.0+1",
+        "1.0.0+build.1",
+    )
```
tomschr: That test uses an `in` assertion agains a tuple of possible outcomes. However, for each specific input (`""` or `None`), the resulting version string is deterministic. This weak assertion could allow incorrect behavior to pass the test.


```suggestion
@pytest.mark.parametrize(
    "token, expected_prerelease, expected_build",
    [
        ("", "1.0.0-1", "1.0.0+1"),
        (None, "1.0.0-rc.1", "1.0.0+build.1"),
    ],                                                                                                                              
    ids=["empty-string", "none"],
)
def test_bump_methods_still_accept_empty_and_none(token, expected_prerelease, expected_build):
    """Empty string and None are documented and must keep their meaning."""
    assert str(Version.parse("1.0.0").bump_prerelease(token)) == expected_prerelease
    assert str(Version.parse("1.0.0").bump_build(token)) == expected_build
```

src/semver/version.py
Str0k: Good point, done in 3cf82bd. Each identifier pattern is now defined once (`_PRERELEASE_IDENTIFIER_REGEX`, `_BUILD_IDENTIFIER_REGEX`) and interpolated into `_REGEX_TEMPLATE`; the compiled patterns used by the validation are built from the same strings. I kept the compiled names since the validation already uses them — happy to rename if you prefer.

To check the refactor doesn't change parsing, I compared the old and new regexes on ~100k generated inputs (matches and captured groups): no differences. The suite passes locally (439 tests).

tests/test_bump_token_validation.py
Str0k: Agreed, applied your suggestion in 3cf82bd — the test now checks the exact result for `""` and `None`.

src/semver/version.py
```diff
@@ -110,25 +118,43 @@ class Version:
                 ){opt_patch}
             ){opt_minor}
             (?:-(?P<prerelease>
-                (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
-                (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
+                (?:{prerelease})
+                (?:\.(?:{prerelease}))*
```
tomschr: Isn't it easier to just add `_PRERELEASE_IDENTIFIER.pattern`? 🤔 Something like this:


```suggestion
                (?:{_PRERELEASE_IDENTIFIER.pattern})
                (?:\.(?:{_PRERELEASE_IDENTIFIER.pattern}))*
```

src/semver/version.py
```diff
@@ -110,25 +118,43 @@ class Version:
...
             ))?
             (?:\+(?P<build>
-                [0-9a-zA-Z-]+
-                (?:\.[0-9a-zA-Z-]+)*
+                (?:{build})
+                (?:\.(?:{build}))*
```
tomschr: Same here:


```suggestion
                (?:{_BUILD_IDENTIFIER.pattern})
                (?:\.(?:{_BUILD_IDENTIFIER.pattern}))*
```

src/semver/version.py
```diff
@@ -110,25 +118,43 @@ class Version:
...
             \Z
         """
     #: Regex for a semver version
     _REGEX: ClassVar[Pattern[str]] = re.compile(
-        _REGEX_TEMPLATE.format(opt_patch="", opt_minor=""),
+        _REGEX_TEMPLATE.format(
+            opt_patch="",
+            opt_minor="",
+            prerelease=_PRERELEASE_IDENTIFIER_REGEX,
+            build=_BUILD_IDENTIFIER_REGEX,
```
tomschr: After we apply the previous fix, the `prerelease` and `build` would be unnecessary:


```suggestion
        _REGEX_TEMPLATE.format(
            opt_patch="",
            opt_minor="",
```

src/semver/version.py
```diff
@@ -110,25 +118,43 @@ class Version:
...
         re.VERBOSE | re.ASCII,
     )
     #: Regex for a semver version that might be shorter
     _REGEX_OPTIONAL_MINOR_AND_PATCH: ClassVar[Pattern[str]] = re.compile(
-        _REGEX_TEMPLATE.format(opt_patch="?", opt_minor="?"),
+        _REGEX_TEMPLATE.format(
+            opt_patch="?",
+            opt_minor="?",
+            prerelease=_PRERELEASE_IDENTIFIER_REGEX,
+            build=_BUILD_IDENTIFIER_REGEX,
```
tomschr: Here too:

```suggestion
```

src/semver/version.py
```diff
@@ -110,25 +118,43 @@ class Version:
                 ){opt_patch}
             ){opt_minor}
```
tomschr:
```suggestion
] = rf"""
           ^
           (?P<major>0|[1-9]\d*)
           (?:
               \.
               (?P<minor>0|[1-9]\d*)
               (?:
                   \.
                   (?P<patch>0|[1-9]\d*)
               ){{opt_patch}}
           ){{opt_minor}}
```

src/semver/version.py
```diff
@@ -95,6 +95,14 @@ class Version:
...
     #: Regex for number in a prerelease
     _LAST_PRERELEASE: ClassVar[Pattern[str]] = re.compile(r"^(.*\.)?(\d+)$")
+    #: Regex for a single prerelease identifier, shared by the version
+    #: template and the bump token validation
+    _PRERELEASE_IDENTIFIER_REGEX: ClassVar[str] = (
+        r"0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*"
+    )
+    #: Regex for a single build identifier, shared by the version template
+    #: and the bump token validation
+    _BUILD_IDENTIFIER_REGEX: ClassVar[str] = r"[0-9a-zA-Z-]+"
```
tomschr: This is all correct, but later you define `_BUILD_IDENTIFIER`. I think, we can avoid four class variables and reduce it to two. This helps to keep the class clean.

First we define the pattern object:

```suggestion
    _PRERELEASE_IDENTIFIER: ClassVar[Pattern[str]] = re.compile(
        r"0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*"
    )
    #: Regex for a single build identifier, shared by the version template
    #: and the bump token validation
    _BUILD_IDENTIFIER: ClassVar[Pattern[str]] = re.compile(r"[0-9a-zA-Z-]+")
```

When we need the string representation, we can write:

```python
>>> _BUILD_IDENTIFIER.pattern
'[0-9a-zA-Z-]+'
```

src/semver/version.py
```diff
@@ -110,25 +118,43 @@ class Version:
...
         re.VERBOSE | re.ASCII,
     )
+    #: Regex for a single valid build identifier
+    _BUILD_IDENTIFIER: ClassVar[Pattern[str]] = re.compile(
+        _BUILD_IDENTIFIER_REGEX, re.ASCII
+    )
+    #: Regex for a single valid alphanumeric prerelease identifier
+    _PRERELEASE_IDENTIFIER: ClassVar[Pattern[str]] = re.compile(
+        _PRERELEASE_IDENTIFIER_REGEX, re.ASCII
+    )
```
tomschr: As we already defined them with the same name previously, we can remove them and save two variables. 😉 


```suggestion
```

tomschr (changes requested):
Thank you @Str0k for your contribution! 👍

I have some suggestions for you to review. It's about avoiding too many class variables and using the `.pattern` property from a pattern object.

Let me know what you think! Overall, great work!

Str0k:
Thanks @tomschr, that is simpler. Applied in af07f02.

`_REGEX_TEMPLATE` now references `_PRERELEASE_IDENTIFIER.pattern` and `_BUILD_IDENTIFIER.pattern` directly. I removed the two intermediate string patterns, the duplicate compiled definitions, and the corresponding `format()` arguments, so only the two compiled class variables remain.

The full suite passes (438 passed, 1 skipped, 100% coverage). I also compared the old and new versions on 100,028 deterministic inputs for each of the four affected patterns; matches and captured groups were identical.

tomschr:
Thank you @Str0k, that looks great now! :+1: I'm ready to merge it.

